PR #1, opened 2018-10-16
download the predictions every so often

sghiaa:
putting this up there for initial feed back, though i need to add tests and do something with the file we download. 

[We can see archived recent predictions in DB](https://app.asana.com/0/584764604969369/865478552763964)


config/config.exs
```diff
@@ -27,4 +27,19 @@ use Mix.Config
 # Configuration from the imported file will override the ones defined
 # here (which is why it is important to import them last).
 #
-#     import_config "#{Mix.env}.exs"
+
+config :prediction_analyzer, Predictions.Repo,
+  adapter: Ecto.Adapters.Postgres,
+  database: "prediction_analyzer_repo",
+  username: "user",
```
losvedir: Can you use whatever syntax it is that Ecto provides in order to pull the credentials from ENV?

lib/predictions/download.ex
```diff
@@ -0,0 +1,65 @@
...
+            trip_prediction
+            | arrival_time: update["arrival"]["time"],
+              departure_time: update["departure"]["time"],
+              boarding_status: update["boarding_status"],
+              schedule_relationship: update["schedule_relationship"],
+              stop_id: update["stop_id"],
+              stop_sequence: update["stop_sequence"],
+              stops_away: update["stops_away"]
+          }
+          |> Predictions.Repo.insert()
```
losvedir: Does Ecto not provide a bulk insert function?

losvedir:
Could you update the README with how to run the app?

lib/predictions/download.ex
sghiaa: it does, ill reformat this so it works with that.

config/config.exs
sghiaa: yeah

sghiaa:
> Could you update the README with how to run the app?

done

lib/predictions/repo.ex
```diff
@@ -0,0 +1,22 @@
+defmodule Predictions.Repo do
+  use Ecto.Repo, otp_app: :prediction_analyzer
+
+  def init(_, opts) do
+    opts =
+      opts
+      |> Keyword.put(
+        :username,
```
losvedir: I think we'll also have to configure the port and dbname. Can you just use the database url which should contain everything we need as far as credentials and configuration go?

config/dev.exs
```diff
@@ -51,8 +57,5 @@ config :phoenix, :stacktrace_depth, 20
 # Configure your database
 config :prediction_analyzer, PredictionAnalyzer.Repo,
   adapter: Ecto.Adapters.Postgres,
-  username: "postgres",
-  password: "postgres",
-  database: "prediction_analyzer_dev",
-  hostname: "localhost",
+  database: "prediction_analyzer_repo",
```
losvedir: I don't think it's necessary here, since you're specifying it in `config.exs`.

lib/predictions/download.ex
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def init(args) do
+    schedule_fetch(self(), 1_000)
+    {:ok, args}
+  end
+
+  def get_predictions() do
+    {aws_requestor, bucket_name, path_name} = get_aws_vars()
+    {:ok, object} = ExAws.S3.get_object(bucket_name, path_name) |> aws_requestor.request()
```
losvedir: Can you add an info logging statement here?

lib/predictions/repo.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Predictions.Repo do
+  use Ecto.Repo, otp_app: :prediction_analyzer
+
+  def init(_, opts) do
+    opts =
+      Keyword.put(
+        opts,
+        :url,
+        System.get_env("DB_URL")
```
losvedir: Can you make this `"DATABASE_URL"` so the environment variable is a little more explicit?

README.md
```diff
@@ -10,6 +10,18 @@ To start your Phoenix server:
...
+## Running the app
+
+In order to run this app locally you will need to set several ENV variables.
+For downloading the Trip updates, you will need an `AWS_ACCESS_KEY_ID` and a `AWS_SECRET_ACCESS_KEY`, as well as `AWS_PREDICTIONS_BUCKET`, and `AWS_PREDICTIONS_PATH`
+
+In order to store them you will need a local postgres database, run the migrations, and set a `DB_URL` in the form:
+`postgresql://<usernam>:<password>@<hostname>:<port>/<database_name>`
+
+Example run command:
+`AWS_PREDICTIONS_BUCKET=<bucket> AWS_PREDICTIONS_PATH=<path> AWS_ACCESS_KEY_ID=<access_id> AWS_SECRET_ACCESS_KEY=<secret_access_id> DB_URL=<url> DB_HOSTNAME=<hostname> iex -S mix`
```
losvedir: Just a note that after you change to `DATABASE_URL` this will need to be updated. Also, I don't think `DB_HOSTNAME` is used?

config/dev.exs
sghiaa: its actually also part of the db_url, so ill remove that.

lib/predictions/download.ex
sghiaa: ok

lib/predictions/repo.ex
sghiaa: sure

README.md
sghiaa: i guess i missed that delete

config/dev.exs
sghiaa: i guess it was necessary in config.exs

PR #5, opened 2018-10-22
make an html table to look at the next 100 predictions by arrival time

sghiaa:
[[3] A user can see recent predictions](https://app.asana.com/0/584764604969369/872531721286022)

this encompasses all the subtasks sorry gabe :/

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -13,6 +13,7 @@ config :logger, level: :warn
 config :prediction_analyzer, Predictions.Repo,
   username: System.get_env("DATABASE_POSTGRESQL_USERNAME") || "postgres",
   password: System.get_env("DATABASE_POSTGRESQL_PASSWORD") || "postgres",
+  adapter: Ecto.Adapters.Postgres,
   database: "prediction_analyzer_test",
   hostname: "localhost",
   pool: Ecto.Adapters.SQL.Sandbox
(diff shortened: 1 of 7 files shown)


config/test.exs
```diff
@@ -13,6 +13,7 @@ config :logger, level: :warn
 config :prediction_analyzer, Predictions.Repo,
   username: System.get_env("DATABASE_POSTGRESQL_USERNAME") || "postgres",
   password: System.get_env("DATABASE_POSTGRESQL_PASSWORD") || "postgres",
+  adapter: Ecto.Adapters.Postgres,
```
losvedir: Is this necessary? I think this is specified in `config.exs`.

sghiaa: i was getting it complaining about this, but its possible taht i fixed the complaints in another way - ill remove this if it allows me to

lib/prediction_analyzer_web/templates/predictions/index.html.eex
```diff
@@ -0,0 +1,32 @@
+<div>
+  HI THERE
```
losvedir: Heh. Maybe this should be an `h1` that says "Recent Predictions" or something like that?

sghiaa: lol i though i deleted that

PR #11, opened 2018-10-25
migration for vehicle_event_id

sghiaa:
[(.5) Migration to add vehicle_event_id to predictions table](https://app.asana.com/0/584764604969369/878909118663894)

add vehicle_event_id to predictions.

### DIFF
--- a/priv/repo/migrations/20181025152446_add_vehicle_event_id_to_predictions.exs
+++ b/priv/repo/migrations/20181025152446_add_vehicle_event_id_to_predictions.exs
@@ -3,7 +3,7 @@ defmodule PredictionAnalyzer.Repo.Migrations.AddVehicleEventIdToPredictions do
 
   def change do
     alter table("predictions") do
-      add(:vehicle_event_id, references("vehicle_events"))
+      add(:vehicle_event_id, references("vehicle_events"), on_delete: :delete_all)
     end
   end
 end
--- a/priv/repo/migrations/20181025152446_add_vehicle_event_id_to_predictions.exs
+++ b/priv/repo/migrations/20181025152446_add_vehicle_event_id_to_predictions.exs
@@ -3,7 +3,7 @@ defmodule PredictionAnalyzer.Repo.Migrations.AddVehicleEventIdToPredictions do
 
   def change do
     alter table("predictions") do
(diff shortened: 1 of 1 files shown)


priv/repo/migrations/20181025152446_add_vehicle_event_id_to_predictions.exs
```diff
@@ -0,0 +1,9 @@
+defmodule PredictionAnalyzer.Repo.Migrations.AddVehicleEventIdToPredictions do
+  use Ecto.Migration
+
+  def change do
+    alter table("predictions") do
+      add(:vehicle_event_id, :id)
```
losvedir: Do ecto migrations let you specify somehow that this is a foreign key?

sghiaa: i think you would change this in the schema, which will come as a later taks in this ticket.

losvedir: I mean, an FK constraint at the DB level, which you would add when create the column in the DB.

sghiaa: oh, good point.  that does seem possible; fixed

priv/repo/migrations/20181025152446_add_vehicle_event_id_to_predictions.exs
```diff
@@ -0,0 +1,9 @@
+defmodule PredictionAnalyzer.Repo.Migrations.AddVehicleEventIdToPredictions do
+  use Ecto.Migration
+
+  def change do
+    alter table("predictions") do
+      add(:vehicle_event_id, references("vehicle_events"))
```
losvedir: Hm, thinking about it more, we're going to want to prune these tables in the future, and so we'll have to manage the foreign key wrt to deleting. I think, since we want to delete old `vehicle_events` and `predictions`, if the referenced vehicle event is deleted, we should allow that and just delete all the predictions that point to it as well.

So then can you update this to be `add(:vehicle_event_id, references("vehicle_events"), on_delete: :delete_all)`?

PR #12, opened 2018-10-25
Watch for vehicle movements and save arrivals/departures to DB

losvedir:
Asana [(2) Logic in GenServer to compare new predictions to old, and log events](https://app.asana.com/0/584764604969369/876349910542899) and [(1) Update DB with vehicle events](https://app.asana.com/0/584764604969369/876349910542904).

This adds the `VehicleEvent` schema to query the `vehicle_events` table. It updates the logic of the VehiclePositions.Tracker to compare the old positions to the new, with a new `Comparator` module. If a given vehicle has changed its status to or from "STOPPED_AT" with the same or different stop_id, then we either create a new vehicle_event record for the arrival, or attempt to update an existing record with the departure time.

### DIFF
--- a/lib/prediction_analyzer/vehicle_positions/comparator.ex
+++ b/lib/prediction_analyzer/vehicle_positions/comparator.ex
@@ -1,0 +1,100 @@
+defmodule PredictionAnalyzer.VehiclePositions.Comparator do
+  @moduledoc """
+  Compare the new set of vehicles to the old ones, to determine
+  which ones have arrived at or departed from a station.
+  """
+
+  require Logger
+  alias PredictionAnalyzer.VehiclePositions.Tracker
+  alias PredictionAnalyzer.VehiclePositions.Vehicle
+  alias PredictionAnalyzer.VehicleEvents.VehicleEvent
+  alias PredictionAnalyzer.Repo
+  import Ecto.Query, only: [from: 2]
+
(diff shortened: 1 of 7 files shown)


lib/prediction_analyzer/vehicle_positions/comparator.ex
```diff
@@ -0,0 +1,100 @@
...
+      where: ve.vehicle_id == ^(vehicle.id)
+        and ve.stop_id == ^(vehicle.stop_id)
+        and is_nil(ve.departure_time)
+        and ve.arrival_time > ^(:os.system_time(:second) - 60*30),
+      update: [set: [departure_time: ^(vehicle.timestamp)]]
+    )
+    |> Repo.update_all([])
+    |> case do
+      {0, _} ->
+        Logger.warn("Tried to update departure time, but no arrival for #{vehicle.label}")
```
sghiaa: wont this happen when arriving at the first stop of a trip?

losvedir: I don't think so. Maybe the first trip of the day for that vehicle... but even then don't we do `INCOMING_AT` and then `STOPPED_AT` in VehiclePositions (not TripUpdates) for that vehicle?

sghiaa: im not sure, maybe youre right

losvedir: Well we can monitor the warnings in the log to see.

PR #13, opened 2018-10-25
Associate new vehicle events with their predictions

losvedir:
Asana tasks:
* [(4) When vehicle_event is created, update predictions table relevant rows with that new ID](https://app.asana.com/0/584764604969369/878909118663891)
* [(.5) Log a warning if no associated prediction](https://app.asana.com/0/584764604969369/878889370171877)

Uncovered a bug with how we were saving predictions: the `trip_id` isn't the GTFS entity ID, but rather buried in entity -> trip_update -> trip -> trip_id.

This PR associates new vehicle_events with their corresponding predictions, according to the algorithm in the asana ticket: matching stop_id and trip_id, with a time constraint. The asana ticket says "today", but I limited it to "arrival_time" OR "departure_time" predicted within 2 hours of now. (i.e. can't be off by more than 2 hours).

Ideally that time constraint will be against the prediction _file_ rather than this arrival OR departure thing, but that column is coming in a later ticket. I've left a note on that ticket saying it should update this where clause once we have that column.

### DIFF
--- a/lib/prediction_analyzer/predictions/download.ex
+++ b/lib/prediction_analyzer/predictions/download.ex
@@ -43,7 +43,7 @@ defmodule PredictionAnalyzer.Predictions.Download do
     predictions =
       Enum.flat_map(entities, fn prediction ->
         trip_prediction = %{
-          trip_id: prediction["id"],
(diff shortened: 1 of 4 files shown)


sghiaa:
oh i see thats covered in the next one.

PR #14, opened 2018-10-25
Ash view vehicles

sghiaa:
[(.25) Phoenix route for /vehicle_events](https://app.asana.com/0/584764604969369/876349910542900)
[(1) Phoenix controller that queries DB](https://app.asana.com/0/584764604969369/876349910542901)
[(1) Phoenix view/template to list in HTML table](https://app.asana.com/0/584764604969369/876349910542902)

heres the router controller and template/view


losvedir:
Can you update the description to point to the relevant tasks rather than the parent? There's a lot of PRs for that one story, and if we're spelunking in the future from a `git blame` or something, it'd be good to easily see what tasks this addresses.

lib/prediction_analyzer_web/controllers/vehicle_events_controller.ex
```diff
@@ -0,0 +1,12 @@
+defmodule PredictionAnalyzerWeb.VehicleEventsController do
+  use PredictionAnalyzerWeb, :controller
+  import Ecto.Query, only: [from: 2]
+  alias PredictionAnalyzer.VehicleEvents.VehicleEvent
+
+  def index(conn, _params) do
+    query = from(vp in VehicleEvent, order_by: [:arrival_time, :departure_time], limit: 100)
```
losvedir: I think we can order by just `:arrival_time`. The second column is only for a time breaker, and I don't know how many vehicles will be arriving at the exact same second.

sghiaa: i suppose thats true

lib/prediction_analyzer_web/templates/vehicle_events/index.html.eex
```diff
@@ -0,0 +1,29 @@
+<h1>Vehicle Events</h1>
+
+<table>
+  <tr>
+      <td>vehicle_id</td>
```
losvedir: I think these should be `<th>`.

test/prediction_analyzer_web/controllers/vehicle_events_controller_test.exs
```diff
@@ -0,0 +1,32 @@
...
+    assert response =~ "vehicle_id"
+    assert response =~ "vehicle_label"
+    assert response =~ "is_deleted"
+    assert response =~ "route_id"
+    assert response =~ "direction_id"
+    assert response =~ "trip_id"
+    assert response =~ "stop_id"
+    assert response =~ "arrival_time"
+    assert response =~ "departure_time"
+  end
```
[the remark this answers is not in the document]
losvedir: When you do `async: true` it checks out the SQL Sandbox adapter in `:shared` mode, which I guess is necessary when multiple things in the same process need it. I wasn't sure if the controller fetch would see the DB changes or if it'd be in its own transaction, but it looks like it can see it. As it is, the DB transaction I think should be rolled back after this test, so I don't think it'd be necessary with other tests.

PR #17, opened 2018-10-26
Add PredictionAccuracy schema

losvedir:
Related to asana ticket [[5] We can see # of accurate and # of total predictions per stop per hour per bin](https://app.asana.com/0/584764604969369/878889370171875), but not any of its tasks in particular. Based on learnings from last sprint, we found that task inter-dependencies complicated working on the story in parallel. One of the hangups was the struct related to the table, which many of the tasks used. This PR introduces the struct in the hope of facilitating parallel work with fewer conflicts.


losvedir:
Gonna merge anyway because the tests pass locally and the semaphore failures are from Semaphore being configured for a different branch.

PR #18, opened 2018-10-26
Codecov

sghiaa:
no branch but we need this


losvedir:
Hmm, can you split up the CodeCov and formatter changes? The format diff might be large, and it'd be some comfort to know that it came from `mix format` and nothing else (well, plus adding `.formatter.exs` to the repo). Also, might be worth getting the formatter in _before_ CodeCov so we're not dinged on reformatting code... ;-)

semaphore/static_analysis.sh
```diff
@@ -2,6 +2,7 @@
 set -e
 set -x
 
-mix format --check-formatted
+MIX_ENV=test mix coveralls.json
+bash <(curl -s https://codecov.io/bash) -t $PREDICTION_ANALYZER_CODECOV_TOKEN -r mbta/prediction_analyzer
 
-mix test
```
losvedir: Tests aren't just for nerds!

losvedir: Ah.

PR #20, opened 2018-10-26
Formatter

sghiaa:
formatting


semaphore/static_analysis.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -e
+set -x
+
+mix format mix.exs "lib/**/*.{ex,exs}" "test/**/*.{ex,exs}" --check-formatted
```
losvedir: Since there's a `.formatter.exs` included, can you make this just `mix format --check-formatted`, so that if we decide to change something about the formatting in the future, we don't have to remember to also change here?

sghiaa: ok

losvedir:
Hmm... did you remove `mix test` from the semaphore job?

sghiaa:
tests are for nerds

losvedir:
Hmm, so actually rather than cramming everything into this `static_analysis.sh` file, why not make the different tests different semaphore jobs so they can run in parallel? `mix format --check-formatted` could be its own job, `mix test` another. See how [the API](https://semaphoreci.com/mbta/api/settings) does it. You don't even need a `static_analysis.sh` file in this PR, really.

sghiaa:
this is the style of rtr and i think we took it from dotcom.  if we want to do something different, thats fine too i guess though

losvedir:
I know the benefit of consistency, but I think its outweighed by having all our static analyses happen in parallel, to bring the build time down.

sghiaa:
there is also the benefit of having it in ci, but doesnt matter to me.  whatever we want to do

losvedir:
Eh, that's a good point. Okay, let's keep it this way then.

PR #25, opened 2018-10-29
add environment to predictions:

sghiaa:
[(1) Update Predictions.Download to download from prod and dev-green and add to DB](https://app.asana.com/0/584764604969369/884753442917085)

codecov seems to be giving me a 500, but ill add some tests

### DIFF
--- a/lib/prediction_analyzer/application.ex
+++ b/lib/prediction_analyzer/application.ex
@@ -48,6 +48,12 @@ defmodule PredictionAnalyzer.Application do
 
   defp set_runtime_config do
     Config.update_env(:aws_predictions_url, System.get_env("AWS_PREDICTIONS_URL"))
+
+    Config.update_env(
+      :dev_green_aws_predictions_url,
+      System.get_env("DEV_GREEN_AWS_PREDICTIONS_URL")
+    )
+
     Config.update_env(:aws_vehicle_positions_url, System.get_env("AWS_VEHICLE_POSITIONS_URL"))
   end
 end
--- a/lib/prediction_analyzer/predictions/download.ex
+++ b/lib/prediction_analyzer/predictions/download.ex
@@ -9,41 +9,65 @@ defmodule PredictionAnalyzer.Predictions.Download do
   end
 
   def init(args) do
-    schedule_fetch(self(), 1_000)
+    schedule_prod_fetch(self(), 1_000)
+    schedule_dev_green_fetch(self(), 31_000)
     {:ok, args}
   end
 
-  def get_predictions() do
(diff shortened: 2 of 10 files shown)


test/prediction_analyzer/prediction_accuracy/prediction_accuracy_test.exs
```diff
@@ -2,6 +2,10 @@ defmodule PredictionAnalyzer.PredictionAccuracy.PredictionAccuracyTest do
   use ExUnit.Case, async: true
   alias PredictionAnalyzer.PredictionAccuracy.PredictionAccuracy
 
+  setup do
```
losvedir: This test doesn't interact with the DB so this setup block shouldn't be needed.

sghiaa: hmm at the time i added that it seemed to fix one of those database errors, but youre right; upon removal no new ones reappeared.

lib/prediction_analyzer/application.ex
```diff
@@ -48,6 +48,12 @@ defmodule PredictionAnalyzer.Application do
 
   defp set_runtime_config do
     Config.update_env(:aws_predictions_url, System.get_env("AWS_PREDICTIONS_URL"))
+
+    Config.update_env(
+      :dev_green_aws_predictions_url,
+      System.get_env("DEV_GREEN_AWS_PREDICTIONS_URL")
```
losvedir: Can you update the README about starting the app with this ENV var?

lib/prediction_analyzer/predictions/download.ex
```diff
@@ -9,41 +9,65 @@ defmodule PredictionAnalyzer.Predictions.Download do
   end
 
   def init(args) do
-    schedule_fetch(self(), 1_000)
+    schedule_prod_fetch(self(), 1_000)
+    schedule_dev_green_fetch(self(), 31_000)
```
losvedir: I do like the idea of the offset, since it will help from overwhelming the DB in a burst, but I worry that the predictions might be coincidentally different enough that the accuracy isn't identical between prod and dev-green even when they're running the same code. What about something smaller like `11_000`?

lib/prediction_analyzer/application.ex
sghiaa: ah, right, good point.

lib/prediction_analyzer/predictions/download.ex
sghiaa: sure, that is probably just as fine.

PR #26, opened 2018-10-30
Hourly calculation of prediction accuracy

losvedir:
Asana: [(6) New GenServer to run once an hour to calculate aggregates](https://app.asana.com/0/584764604969369/878909118663898).

Introduces a new worker GenServer into our Application tree called `PredictionAccuracy.Aggregator`. It runs once an hour at the top of the hour executing a few SQL `INSERT INTO prediction_accuracy .... (SELECT FROM ....)` queries.


lib/prediction_analyzer/prediction_accuracy/query.ex
```diff
@@ -0,0 +1,95 @@
...
+        $4 AS bin,
+        COUNT(*) AS num_predictions,
+        SUM(
+          CASE
+            WHEN
+              ve.#{column_name} - p.#{column_name} > $7
+              AND ve.#{column_name} - p.#{column_name} < $8 THEN 1
+            ELSE 0
+          END
+      ) AS num_accurate_predictions
```
dmaltzan: I think line 82 should be indented further... it goes with the SUM, right?

losvedir: /gif right you are, ken

lib/prediction_analyzer/prediction_accuracy/query.ex
```diff
@@ -0,0 +1,95 @@
...
+        bin_error_min,
+        bin_error_max,
+        min_unix,
+        max_unix
+      ]
+    )
+  end
+
+  defp query_template(arrival_departure) do
+    column_name =
```
dmaltzan: can we rename `column_name` to something like `arrival_or_departure`... that's a little long but something like that

PR #28, opened 2018-10-31
Gjd ash dev green preds for codecov

losvedir:
Just opening a new PR with the same commits as #25 to see if that tricks CodeCov into deciding to run.


losvedir:
Nope, didn't work.

PR #30, opened 2018-11-01
Ash dev green vehicles

sghiaa:
[(1) Update VehiclePositions.Tracker to download from mbta-gtfs-s3 and mbta-gtfs-s3-dev-green](https://app.asana.com/0/584764604969369/884753442917084)

### DIFF
--- a/lib/prediction_analyzer/application.ex
+++ b/lib/prediction_analyzer/application.ex
@@ -60,6 +60,11 @@ defmodule PredictionAnalyzer.Application do
     Config.update_env(:aws_vehicle_positions_url, System.get_env("AWS_VEHICLE_POSITIONS_URL"))
 
     Config.update_env(
+      :dev_green_aws_predictions_url,
+      System.get_env("DEV_GREEN_AWS_PREDICTIONS_URL")
+    )
+
+    Config.update_env(
       :dev_green_aws_vehicle_positions_url,
       System.get_env("DEV_GREEN_AWS_VEHICLE_POSITIONS_URL")
     )
--- a/lib/prediction_analyzer/application.ex
+++ b/lib/prediction_analyzer/application.ex
@@ -15,13 +15,7 @@ defmodule PredictionAnalyzer.Application do
       supervisor(PredictionAnalyzer.Repo, []),
       worker(
         PredictionAnalyzer.VehiclePositions.Tracker,
-        [
-          [
-            environment: "dev-green",
-            aws_predictions_url:
-              "https://s3.amazonaws.com/mbta-gtfs-s3-dev-green/rtr/VehiclePositions_enhanced.json"
-          ]
-        ],
+        [[environment: "dev-green"]],
         id: DevGreenVehiclePositionsTracker
       ),
       worker(
@@ -63,12 +57,11 @@ defmodule PredictionAnalyzer.Application do
 
   defp set_runtime_config do
     Config.update_env(:aws_predictions_url, System.get_env("AWS_PREDICTIONS_URL"))
+    Config.update_env(:aws_vehicle_positions_url, System.get_env("AWS_VEHICLE_POSITIONS_URL"))
 
     Config.update_env(
-      :dev_green_aws_predictions_url,
-      System.get_env("DEV_GREEN_AWS_PREDICTIONS_URL")
+      :dev_green_aws_vehicle_positions_url,
+      System.get_env("DEV_GREEN_AWS_VEHICLE_POSITIONS_URL")
     )
-
-    Config.update_env(:aws_vehicle_positions_url, System.get_env("AWS_VEHICLE_POSITIONS_URL"))
   end
 end
--- a/lib/prediction_analyzer/application.ex
+++ b/lib/prediction_analyzer/application.ex
@@ -13,7 +13,22 @@ defmodule PredictionAnalyzer.Application do
     children = [
       # Start the Ecto repository
       supervisor(PredictionAnalyzer.Repo, []),
-      worker(PredictionAnalyzer.VehiclePositions.Tracker, []),
+      worker(
+        PredictionAnalyzer.VehiclePositions.Tracker,
+        [
+          [
+            environment: "dev-green",
+            aws_predictions_url:
+              "https://s3.amazonaws.com/mbta-gtfs-s3-dev-green/rtr/VehiclePositions_enhanced.json"
+          ]
+        ],
+        id: DevGreenVehiclePositionsTracker
+      ),
+      worker(
+        PredictionAnalyzer.VehiclePositions.Tracker,
+        [[environment: "prod"]],
+        id: ProdVehiclePositionsTracker
+      ),
       worker(PredictionAnalyzer.Predictions.Download, [
         [name: PredictionAnalyzer.Predictions.Download]
       ]),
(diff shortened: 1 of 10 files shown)


lib/prediction_analyzer/application.ex
```diff
@@ -13,7 +13,22 @@ defmodule PredictionAnalyzer.Application do
...
       # Start the Ecto repository
       supervisor(PredictionAnalyzer.Repo, []),
-      worker(PredictionAnalyzer.VehiclePositions.Tracker, []),
+      worker(
+        PredictionAnalyzer.VehiclePositions.Tracker,
+        [
+          [
+            environment: "dev-green",
+            aws_predictions_url:
+              "https://s3.amazonaws.com/mbta-gtfs-s3-dev-green/rtr/VehiclePositions_enhanced.json"
```
losvedir: It feels weird to me that you have to specify this here, but the prod one isn't specified. Does that mean this doesn't need an ENV? If so, it's the only AWS URL that doesn't. I think it should work consistently with the way prod is configured.

lib/prediction_analyzer/vehicle_positions/tracker.ex
```diff
@@ -9,23 +9,31 @@ defmodule PredictionAnalyzer.VehiclePositions.Tracker do
 
   @type t :: %{
           http_fetcher: module(),
+          environment: String.t(),
           aws_vehicle_positions_url: String.t(),
           vehicles: vehicle_map()
         }
 
-  def start_link(opts \\ []) do
+  def start_link(opts \\ [], args) do
```
losvedir: What's the difference between `opts` and `args` and why is `:aws_vehicle_positions_url` in one and `:environment` in the other?

losvedir: I guess it was a lazy Socratic method question. I know the difference: one is for starting up the GenServer, and the other is for the GenServer itself to use. So, e.g.,; you'll typically have `GenServer.start_link(__MODULE__, args, opts)`, where the `opts` would be something like `name: Foo` (meaning, this GenServer is known to the system by the name `Foo`), and the args, are what the GenServer itself receives in the `init(args)` function.

So, I would say `environment` and `aws_vehicle_positions_url` are _both_ `args`, in that it's something that the GenServer should know about and store.  Since at the bottom of the function you call `GenServer.start_link(__MODULE__, initial_state)` and don't even pass in the options to that call, it looks like you don't need the `opts` at all!

On the other hand, and I haven't fully looked at this PR, if your goal is to start up two different GenServers using the same module, then you probably will need `opts`, so that the application start callback can specify two different names for each of them.

sghiaa: we could name them, i suppose.  i added ids for them because thats how they are differentiated internally i think.

lib/prediction_analyzer/application.ex
```diff
@@ -48,12 +57,11 @@ defmodule PredictionAnalyzer.Application do
 
   defp set_runtime_config do
     Config.update_env(:aws_predictions_url, System.get_env("AWS_PREDICTIONS_URL"))
+    Config.update_env(:aws_vehicle_positions_url, System.get_env("AWS_VEHICLE_POSITIONS_URL"))
 
     Config.update_env(
-      :dev_green_aws_predictions_url,
-      System.get_env("DEV_GREEN_AWS_PREDICTIONS_URL")
+      :dev_green_aws_vehicle_positions_url,
+      System.get_env("DEV_GREEN_AWS_VEHICLE_POSITIONS_URL")
```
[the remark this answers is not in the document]
sghiaa: im not sure what you mean by this. i see it only once, here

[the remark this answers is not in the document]
losvedir: Ah, yep. Not sure what I was seeing... Should there be `DEV_GREEN_AWS_PREDICTIONS_URL` somewhere, though? Or is this branch out of date? Can you rebase, if so, to bring in the most recent stuff. I worry there might be some conflicts or something, so I want to review after that.

lib/prediction_analyzer/vehicle_positions/comparator.ex
```diff
@@ -88,6 +88,7 @@ defmodule PredictionAnalyzer.VehiclePositions.Comparator do
   @spec vehicle_params(Vehicle.t()) :: map()
   defp vehicle_params(vehicle) do
     %{
+      environment: vehicle.environment,
```
losvedir: In this file, I think you'll also need to update `associate_vehicle_event_with_predictions`, right? Otherwise you'll have dev green vehicles updating prod predictions and vice versa.

sghiaa: yeah, we do. i interpreted that as the next task, but i can do that in this

lib/prediction_analyzer/application.ex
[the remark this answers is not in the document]
sghiaa: hmm, yeah there should.  i wonder if that got lost in a rebase earlier because im allegedly up to date.  ill put it back in.

lib/prediction_analyzer/vehicle_positions/comparator.ex
sghiaa: done

lib/prediction_analyzer/vehicle_positions/tracker.ex
```diff
@@ -9,23 +9,27 @@ defmodule PredictionAnalyzer.VehiclePositions.Tracker do
...
       Keyword.get(
-        opts,
-        :aws_vehicle_positions_url,
-        Application.get_env(:prediction_analyzer, :aws_vehicle_positions_url)
+        args,
+        :environment
       )
 
+    aws_vehicle_positions_url =
+      get_env_vehicle_positions_url(environment) || args[:aws_vehicle_positions_url]
```
losvedir: I don't see any way for `get_env_vehicle_positions_url` to return `nil` so I don't think the `||` is necessary.

lib/prediction_analyzer/vehicle_positions/tracker.ex
```diff
@@ -9,23 +9,27 @@ defmodule PredictionAnalyzer.VehiclePositions.Tracker do
...
+
     http_fetcher =
-      Keyword.get(opts, :http_fetcher, Application.get_env(:prediction_analyzer, :http_fetcher))
+      Keyword.get(args, :http_fetcher, Application.get_env(:prediction_analyzer, :http_fetcher))
 
     initial_state = %{
       aws_vehicle_positions_url: aws_vehicle_positions_url,
+      environment: args[:environment],
```
losvedir: Maybe just use `environment` here since you extracted it above?

lib/prediction_analyzer/vehicle_positions/tracker.ex
sghiaa: in the test this was getting nil because we are passing it in.

lib/prediction_analyzer/vehicle_positions/tracker.ex
sghiaa: good idea

lib/prediction_analyzer/vehicle_positions/tracker.ex
losvedir: Ah, I see.

PR #31, opened 2018-11-01
Add filtering of prediction accuracy

losvedir:
Asana: [[3] We can drill down into the accuracy metrics by attribute](https://app.asana.com/0/584764604969369/881339877158243).

Adds a form on the page that lets you filter by the specified attributes. The rows on the page reflect the filters, as does the accuracy summary on top. The form fields maintain their state based on the parameters in the URL.

screen shot 2018-11-01 at 1 02 36 pm

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -8,7 +8,7 @@ use Mix.Config
 # Configures the endpoint
 config :prediction_analyzer, PredictionAnalyzerWeb.Endpoint,
   url: [host: "localhost"],
-  secret_key_base: "local_secret_key_base",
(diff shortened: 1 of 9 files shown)


config/config.exs
```diff
@@ -8,7 +8,7 @@ use Mix.Config
 # Configures the endpoint
 config :prediction_analyzer, PredictionAnalyzerWeb.Endpoint,
   url: [host: "localhost"],
-  secret_key_base: "local_secret_key_base",
+  secret_key_base: "local_secret_key_base_at_least_64_bytes_________________________________",
```
sghiaa: lol

PR #34, opened 2018-11-02
Ash aggregate dev green

sghiaa:
[(1) Update PredictionAccuracy.Aggregator to do the calculations for prod and dev-green](https://app.asana.com/0/584764604969369/884753442917086)
[(2) Update PredictionAccuracy.query to filter by env, and controller to call it once for each env. Add second accuracy to page. Update table to have ENV column and display data from both envs](https://app.asana.com/0/584764604969369/884753442917087)

this covers both of these fcause they didnt realistically separate well

screenshot:

screen shot 2018-11-02 at 4 35 51 pm

### DIFF
--- a/lib/prediction_analyzer/prediction_accuracy/aggregator.ex
+++ b/lib/prediction_analyzer/prediction_accuracy/aggregator.ex
@@ -20,32 +20,56 @@ defmodule PredictionAnalyzer.PredictionAccuracy.Aggregator do
       :timer.tc(fn ->
         current_time = Timex.local()
 
-        Enum.each(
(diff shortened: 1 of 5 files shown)


losvedir:
Can you attach a screenshot of the /accuracy page?

losvedir:
Hmm, it seems the spacing is broken. The current page has the filters in the sidebar and the column to the right. (See [here](http://prediction-analyzer.mbtace.com).) Something about this has pushed the table data to under the filters.

sghiaa:
screen shot 2018-11-05 at 11 36 15 am

PR #35, opened 2018-11-02
Add Pruner process to delete old data

losvedir:
Asana: [(4) New GenServer that runs every day at 3 am and deletes the old data (and maybe manually runs VACUUM?)](https://app.asana.com/0/584764604969369/884753442917093).

I was reading about the VACUUMing online, and I think AUTO VACUUM might be enough for us. We can watch the metrics to see if our space gets unusable but postgres should be vacuuming itself in a concurrent fashion as necessary. So, I opted not to manually trigger one here.

### DIFF
--- a/lib/prediction_analyzer/pruner.ex
+++ b/lib/prediction_analyzer/pruner.ex
@@ -24,19 +24,24 @@ defmodule PredictionAnalyzer.Pruner do
 
     unix_cutoff = Timex.local() |> Timex.shift(days: -7) |> DateTime.to_unix()
 
-    Repo.delete_all(
-      from(
-        p in Prediction,
-        where: p.file_timestamp < ^unix_cutoff
-      )
-    )
+    {time, _} =
+      :timer.tc(fn ->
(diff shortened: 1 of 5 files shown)


lib/prediction_analyzer/pruner.ex
```diff
@@ -0,0 +1,53 @@
...
+
+  def init([]) do
+    schedule_next_run(self())
+    {:ok, []}
+  end
+
+  def handle_info(:prune, state) do
+    Logger.info("Beginning prune of DB")
+
+    unix_cutoff = Timex.local() |> Timex.shift(days: -7) |> DateTime.to_unix()
```
sghiaa: should we get 3am on the cutoff day?

losvedir: Sure, I guess that's a good thing to add. It shouldn't really be necessary since this is running at 3am, so by shifting back 7 hours we stay at 3am, but might as well add it just to be explicit.

PR #36, opened 2018-11-06
Performance tweaks

losvedir:
This fixes two performance issues I've seen: 

* The Pruner can't finish because the "delete when older than 7 days" times out. I fixed that by adding indexes to those timestamp columns, and split the query into two to better take advantage of them. The table's not _that_ big, so I imagine with these indexes it should now run no problem.

* The logs show the query that associates a prediction with its vehicle_event often takes too long, almost a whole second sometimes. This query runs twice a second (once for prod, once for dev green), so it needs to be faster than that. I changed the query to have a `vehicle_event_id IS NULL` clause (since we only want to update rows that haven't been updated anyway) and added a partial index on predictions for `trip_id` with `vehicle_event_id IS NULL`. That keeps the size of the index down since _most_ predictions should eventually end up with a `vehicle_event_id`. Locally this has dropped these long-running queries down to the 10-20ms range.

Postgres (and Ecto) allows for creating indexes concurrently, that is, without locking the table that's getting the index, but to allow it to happen you have to disable running the migration in a transaction. This isn't ideal, but it's necessary since on my local data it took a minute to create the index on predictions anyway.

The other issue is that since we run migrations as a callback on starting the app, if it takes too long the app will fail. I got around that by running the migrations in a process spawned by the app on start. This isn't great in general, since we don't get feedback if the migrations fail (other than logs), so I plan to revert this particular change once this code has been deployed.

### DIFF
--- a/lib/prediction_analyzer/application.ex
+++ b/lib/prediction_analyzer/application.ex
@@ -48,8 +48,12 @@ defmodule PredictionAnalyzer.Application do
 
     case Supervisor.start_link(supervisors ++ workers, opts) do
       {:ok, _} = success ->
-        Logger.info("Started application, running migrations")
-        Application.get_env(:prediction_analyzer, :migration_task).migrate()
+        spawn(fn ->
+          Logger.info("Started application, running migrations")
+          Application.get_env(:prediction_analyzer, :migration_task).migrate()
+          Logger.info("Finished migrations")
+        end)
+
         success
 
       error ->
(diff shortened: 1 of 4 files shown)


sghiaa:
typically weve had migrations be a separate pr.  are we going with the same pr because this is not a table change, just adding indexes?

losvedir:
Hmm, nah I think you're right. I'll pull out some of the more optional changes into a separate PR.

PR #45, opened 2018-11-19
Download CSV of predictions that went into statistics

losvedir:
Asana: [Click on an hour in the prediction analyzer and see the predictions that went into that as a CSV](https://app.asana.com/0/584764604969369/911114372878693)

screen shot 2018-11-19 at 10 30 18 am

screen shot 2018-11-19 at 10 30 27 am

### DIFF
--- a/lib/prediction_analyzer/prediction_accuracy/aggregator.ex
+++ b/lib/prediction_analyzer/prediction_accuracy/aggregator.ex
@@ -18,7 +18,7 @@ defmodule PredictionAnalyzer.PredictionAccuracy.Aggregator do
 
     {time, _result} =
       :timer.tc(fn ->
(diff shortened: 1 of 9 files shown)


lib/prediction_analyzer/prediction_accuracy/aggregator.ex
```diff
@@ -18,7 +18,7 @@ defmodule PredictionAnalyzer.PredictionAccuracy.Aggregator do
 
     {time, _result} =
       :timer.tc(fn ->
-        current_time = Timex.local()
+        current_time = Timex.now("America/New_York")
```
losvedir: This was a bug for me, but wouldn't affect you guys. The aggregator was using "local" time, which meant Central for me.

PR #47, opened 2018-11-30
add vehicle id and index to predictions

sghiaa:
[(1) Add predictions.vehicle_id column and index (making app start up migrations async, if necessary).](https://app.asana.com/0/584764604969369/929449096386419)


losvedir:
The index should be created concurrently, unless postgres can create an index for a mostly NULL column quickly somehow. (See my other PRs that add indexes for how). Per the ticket I was going to say you also have to make the app start up migrations async, but it looks like [they already are](https://github.com/mbta/prediction_analyzer/blob/25610ebdfe633a2577807a9de375877f518b40f9/lib/prediction_analyzer/application.ex#L51). So we should revert that after this deploy.

PR #48, opened 2018-12-03
use vehicle id instead of trip id in the comparator:

sghiaa:
[(2) Update TripUpdates DL logic to save vehicle_id. Update Comparator to use vehicle_id rather than trip_id. Also consider when vehicle departs station.](https://app.asana.com/0/584764604969369/929449096386420)

### DIFF
--- a/lib/prediction_analyzer/predictions/prediction.ex
+++ b/lib/prediction_analyzer/predictions/prediction.ex
@@ -6,6 +6,7 @@ defmodule PredictionAnalyzer.Predictions.Prediction do
     field(:file_timestamp, :integer)
     field(:environment, :string)
     field(:trip_id, :string)
+    field(:vehicle_id, :string)
     field(:is_deleted, :boolean)
     field(:delay, :integer)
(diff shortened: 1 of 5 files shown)


lib/prediction_analyzer/predictions/prediction.ex
```diff
@@ -31,6 +32,7 @@ defmodule PredictionAnalyzer.Predictions.Prediction do
         "stop_sequence,",
         "stops_away,",
         "trip_id,",
+        "trip_vehicle_id,",
```
dmaltzan: seems fine, just curious - why `trip_vehicle_id` instead of `vehicle_id`?

sghiaa: theres a vehicle_id from the vehicles table later in this.  i wasnt sure if we wanted to get rid of that since it exists already, so i decided to distinguish this one.

dmaltzan: makes sense

PR #50, opened 2018-12-14
Do some rudimentary styling of the page

losvedir:
Asana: [[3] Style prediction analyzer somewhat](https://app.asana.com/0/584764604969369/931950192635050).

Mockup is [here](https://projects.invisionapp.com/share/CVOX8S7YK8Q#/screens/329214426).

screen shot 2018-12-14 at 2 04 01 pm

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -24,17 +24,17 @@
 }
 
 .prediction-analyzer-app {
-  background-color: rgb(2, 36, 65);
+  background-color: #022441;
   color: white;
 }
 
 .accuracy-form {
-  background-color: rgb(14, 67, 110);
+  background-color: #0e4364;
   padding: 1em;
 }
 
 .prod-text {
-  color: rgb(31, 236, 255);
+  color: #1fecff;
 }
 
 .prod-text-header {
@@ -43,7 +43,7 @@
 }
 
 .dev-green-text {
-  color: rgb(199, 67, 240);
+  color: #c743f0;
 }
 
 .dev-green-text-header {
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -1,10 +1,46 @@
 /* This file is for your main application css. */
 
+.mbta-logo {
+  display: inline-block;
+  position: relative;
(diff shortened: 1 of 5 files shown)


assets/css/app.css
```diff
@@ -1,10 +1,46 @@
...
   font-weight: bold;
 }
+
 .accuracy-summary-percent-sign {
   font-size: 46px;
   font-weight: bold;
 }
+
+.prediction-analyzer-app {
+  background-color: rgb(2, 36, 65);
```
sghiaa: im not sure what the stlye guide is for css, but i think the website uses hex codes for colors, as opposed to rgb or names

lib/prediction_analyzer_web/templates/accuracy/index.html.eex
```diff
@@ -45,49 +43,48 @@
   </div>
 
   <div class="col-xs-9">
-    <div class="row">
+    <div class="row accuracy-summary-header">
 
       <div class="col-xs-6">
-        Prod:
+        <span class="prod-text">PROD</span>
```
sghiaa: i think these might want to be bold now, since they seem a little harder to read at this font weight.  or possibly a bigger font size

assets/css/app.css
losvedir: Sure, I can do that. I was using the `Digital Color Meter` app that comes with OS X to measure the colors on the mockup, and it gave them to me in that form. Easy enough to convert. Actually, I think I did for the JS already. I'll change these here.

lib/prediction_analyzer_web/templates/accuracy/index.html.eex
losvedir: Good catch. I realized I should also probably make them all caps with CSS rather than just typing it in that way, too.

losvedir: The mockups have the text pretty small, but I made it bold:

screen shot 2018-12-14 at 2 14 22 pm

PR #52, opened 2018-12-27
Select between hourly or daily accuracies, see results in chart

losvedir:
#### Summary of changes
**Asana Ticket:** [[5] Prediction analyzer allows graph of multiple days](https://app.asana.com/0/584764604969369/923194761406814)

This is the meat of the asana ticket. There's now a form field to switch between "daily" or "hourly" accuracies, which updates the SQL query we do to group appropriately, and render onto the page as a chart.

Subsequent work will make the chart work for daily (it doesn't right now, for some reason which I don't understand), and to move the form element out of the sidebar.

This PR also introduces a script to generate sample data, so you can run `mix ecto.reset` to drop and recreate tables and then fill in two weeks worth of sample accuracy data.

screen shot 2018-12-27 at 3 52 02 pm

#### Reviewer Checklist


assets/css/app.css
```diff
@@ -54,3 +54,25 @@
...
+  color: white;
+  display: inline-block;
+  line-height: 2.5;
+  margin-right: 0.5em;
+  text-align: center;
+  width: 7em;
+}
+
+.button-link-active {
+  background-color: white;
```
lemald: Use hex

losvedir:
Since people were out last week, I just kept adding to this branch, rather than creating a separate one and dealing with two branches cut from each other. Here's what it looks like now:

screen shot 2019-01-02 at 9 11 17 am

PR #55, opened 2019-01-04
Some C3 tweaks

losvedir:
#### Summary of changes
**Asana Ticket:** [[2] Prediction analyzer: label the axes](https://app.asana.com/0/584764604969369/948763837072501) plus some drive bys.

Per the ticket, labels the x- and y- axes. However, when we're in daily view and the dates are rotated, the x-axis label overlapped the x-axis ticks, and I couldn't find a way to prevent it, so I just hid it in that case.

Other tweaks were making the axis text a little bigger, and not rotating the x-axis labels when it's just hours of the day.

screen shot 2019-01-04 at 1 23 39 pm

screen shot 2019-01-04 at 1 23 48 pm

#### Reviewer Checklist

### DIFF
--- a/assets/css/c3.css
+++ b/assets/css/c3.css
@@ -192,3 +192,17 @@
 .c3-tooltip-container {
   color: #000000;
 }
+
+.c3-axis-y-label {
+  fill: #ffffff;
+  font-size: 2em;
+}
+
+.c3-axis-x-label {
+  fill: #ffffff;
+  font-size: 1.5em;
+}
+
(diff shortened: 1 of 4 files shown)


dmaltzan:
Having "Hour of Day" over on the left side feels a little awkward... is there any way to put "Hour of Day" in the middle and the color code over to the right side?

losvedir:
When I put the "Hour of Day" in the middle it overlapped with the legend. But good idea about moving the legend! Yeah, that might be possible... let me investigate.

PR #56, opened 2019-01-04
Prediction Analyzer works when switching to daily in all cases

lemald:
#### Summary of changes
**Asana Ticket:** [Prediction Analyzer works when switching to daily in all cases](https://app.asana.com/0/584764604969369/960766001017225/f)

Don't try to generate the download link for daily charts.

#### Reviewer Checklist

### DIFF
--- a/lib/prediction_analyzer_web/controllers/accuracy_controller.ex
+++ b/lib/prediction_analyzer_web/controllers/accuracy_controller.ex
@@ -51,7 +51,11 @@ defmodule PredictionAnalyzerWeb.AccuracyController do
 
     default_filters = %{
       "chart_range" => "Hourly",
-      "service_date" => Timex.local() |> Date.to_string()
+      "service_date" => Timex.local() |> Date.to_string(),
+      "route_id" => "",
+      "stop_id" => "",
+      "arrival_departure" => "all",
+      "bin" => "All"
     }
 
     filters = Map.merge(default_filters, filters)
--- a/lib/prediction_analyzer_web/controllers/accuracy_controller.ex
+++ b/lib/prediction_analyzer_web/controllers/accuracy_controller.ex
@@ -4,6 +4,7 @@ defmodule PredictionAnalyzerWeb.AccuracyController do
 
   import Ecto.Query, only: [from: 2]
 
+  @spec index(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def index(
         conn,
         params = %{"filters" => %{"chart_range" => chart_range, "service_date" => service_date}}
--- a/lib/prediction_analyzer_web/controllers/accuracy_controller.ex
+++ b/lib/prediction_analyzer_web/controllers/accuracy_controller.ex
@@ -4,6 +4,8 @@ defmodule PredictionAnalyzerWeb.AccuracyController do
 
   import Ecto.Query, only: [from: 2]
 
+  defguard is_binary_present(val) when is_binary(val) and val != ""
+
   @spec index(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def index(conn, %{
         "filters" =>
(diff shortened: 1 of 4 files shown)


lib/prediction_analyzer_web/controllers/accuracy_controller.ex
```diff
@@ -4,8 +4,14 @@ defmodule PredictionAnalyzerWeb.AccuracyController do
 
   import Ecto.Query, only: [from: 2]
 
-  def index(conn, params) do
-    filter_params = params["filters"] || %{}
+  @spec index(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def index(
+        conn,
+        params = %{"filters" => %{"chart_range" => chart_range, "service_date" => service_date}}
```
losvedir: We don't need to capture `params` since it's only used to get to `filter_params`, which you can match on directly:

`%{"filters" => %{ ... } = filter_params}`

lib/prediction_analyzer_web/controllers/accuracy_controller.ex
```diff
@@ -4,8 +4,13 @@ defmodule PredictionAnalyzerWeb.AccuracyController do
 
   import Ecto.Query, only: [from: 2]
 
-  def index(conn, params) do
-    filter_params = params["filters"] || %{}
+  @spec index(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def index(conn, %{
+        "filters" =>
+          %{"chart_range" => chart_range, "service_date" => service_date} = filter_params
```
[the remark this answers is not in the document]
lemald: Good point.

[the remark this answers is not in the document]
lemald: Think this should cover it?
```elixir
    default_filters = %{
      "chart_range" => "Hourly",
      "service_date" => Timex.local() |> Date.to_string(),
      "route_id" => "",
      "stop_id" => "",
      "arrival_departure" => "all",
      "bin" => "All"
    }```

File: lib/prediction_analyzer_web/controllers/accuracy_controller.ex [hunk 3]
```diff
@@ -4,8 +4,22 @@ defmodule PredictionAnalyzerWeb.AccuracyController do
...
+            "service_date" => service_date,
+            "route_id" => route_id,
+            "stop_id" => stop_id,
+            "arrival_departure" => arrival_departure,
+            "bin" => bin
+          } = filter_params
+      })
+      when not is_nil(chart_range) and chart_range != "" and not is_nil(service_date) and
+             service_date != "" and not is_nil(route_id) and not is_nil(stop_id) and
+             not is_nil(arrival_departure) and arrival_departure != "" and not is_nil(bin) and
```
losvedir: "not nil and not equal to the empty string" is a function in the ruby world known as `present?`. I wonder, since we're on Elixir 1.6, if we should take advantage of the new (as of 1.6) `defguard`. Basically, at the top of this module you'd do:

``` ex
defguard is_present(val) when not is_nil(val) and val != ""
```

and then here you'd have `when is_present(chart_range) and is_present(service_date)`.

It seems like a useful guard and we'd probably want to end up putting it somewhere shared, but for now, since this is the only place we're using it, I think it's fine to just define in this module. I also wonder if we should do `is_binary()` (i.e.: "string") rather than `not is_nil`. I like that it's more specific, but then it's not the same as ruby's `present?` method, which might be misleading. I know we have a lot of former rubyists here.

lemald: Good idea! Maybe we could call it like `is_binary_present()` or something? I'm not one of the former Rubyists so I don't have a strong opinion on the naming.

losvedir: Sure, I like that.

paulswartz: If you're trying to check that there's a binary, you can do it in one step with `when byte_size(val) > 0`. Otherwise, you probably want `is_binary(val) and val != ""` to avoid the cases when Phoenix parses the data into a nested map.

lemald: Well I just pushed a commit using the `defguard` approach with `is_binary` and checking for `val !=""`. I'm sort of with @losvedir in finding this approach a tad more readable, but it may in part be that I've only been writing Elixir for a few weeks now and I don't automatically assume that something called `byte_size` is going to make sure that the thing in question is a string.

lemald: Well it looks like Codecov doesn't really play nicely with `defguard` yet, so `byte_size(val) > 0` it is!

PR #59, opened 2019-01-14
migration for not nil vehicle id

sghiaa:
#### Summary of changes
**Asana Ticket:** [Make predictions.vehicle_id NOT NULL after 7 days](https://app.asana.com/0/881264583703207/923338667921775)

here it is, vehicles are supposed to be required now.

#### Reviewer Checklist

### DIFF
--- a/lib/fake_httpoison.ex
+++ b/lib/fake_httpoison.ex
@@ -23,6 +23,9 @@ defmodule FakeHTTPoison do
           "is_deleted" => false,
           "trip_update" => %{
             "delay" => nil,
+            "vehicle" => %{
+              "id" => "vehicle_id"
+            },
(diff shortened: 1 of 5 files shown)


losvedir:
First, I'm not entirely sure it _does_ lock, so can you confirm that question? If it does, then maybe modify seeds.exs locally to add, say, 1 million rows, and then time how long it takes locally? It won't be exactly the same since your local hardware is different from the cloud, and also we have more than that many rows, but just for, like, a rough order of magnitude estimate to go on.

sghiaa:
it seems fast in my testing, and the actual lock time should not be long anyway.  (the scanning to make sure the column can be not null does not require a lock)

PR #61, opened 2019-01-30
Change chart range from select to links

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [[3 extra] Prediction analyzer: format date/time selection more in line with design](https://app.asana.com/0/1101545135637620/990404092148944)

Changed the chart-range "Hourly" vs. "Daily" selection UI from a `select` to a set of links, as pictured in https://projects.invisionapp.com/share/CVOX8S7YK8Q#/screens/329214430

#### Reviewer Checklist

### DIFF
--- a/lib/prediction_analyzer_web/views/accuracy_view.ex
+++ b/lib/prediction_analyzer_web/views/accuracy_view.ex
@@ -61,4 +61,13 @@ defmodule PredictionAnalyzerWeb.AccuracyView do
 
   def button_class(%{}, ""), do: "button-link button-link-active"
   def button_class(%{}, _else), do: "button-link"
+
+  @spec chart_range_class(map(), String.t()) :: String.t()
+  def chart_range_class(%{params: %{"filters" => %{"chart_range" => chart_range}}}, chart_range),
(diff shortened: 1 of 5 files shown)


lib/prediction_analyzer_web/views/accuracy_view.ex
```diff
@@ -61,4 +61,11 @@ defmodule PredictionAnalyzerWeb.AccuracyView do
 
   def button_class(%{}, ""), do: "button-link button-link-active"
   def button_class(%{}, _else), do: "button-link"
+
+  def chart_range_class(%{params: %{"filters" => %{"chart_range" => chart_range}}}, chart_range),
```
phildarnowsky: note to self: add a typespec here

lemald: This function and also `chart_range_id` should probably also have their own tests. Codecov doesn't complain because some other test is causing all of the relevant lines to be called, but there aren't any tests specifically making sure that these functions behave as expected.

phildarnowsky: @lemald this should address your feedback

PR #64, opened 2019-02-27
Clicking on a new route should not reset the form selection

lemald:
#### Summary of changes
**Asana Ticket:** [clicking on a new route should not reset the form selection](https://app.asana.com/0/584764604969369/1109957966379444/f)

Use the same basic method as was used for the hourly / daily buttons.

#### Reviewer Checklist

### DIFF
--- a/assets/js/chart.js
+++ b/assets/js/chart.js
@@ -33,12 +33,11 @@ export function bindFormLinks(accuracyForm) {
   for (var i = 0; i < routeButtonElements.length; i++) {
     routeButtonElements[i].addEventListener('click', function(event) {
       event.preventDefault();
-      routeIdInput.value = event.currentTarget.text;
-      if (this.text === 'All lines') {
+      if (event.currentTarget.text === 'All lines') {
 	routeIdInput.value = '';
       }
       else {
-	var words = this.text.split(' ');
+	var words = event.currentTarget.text.split(' ');
 	routeIdInput.value = words[0];
       }
       accuracyForm.submit();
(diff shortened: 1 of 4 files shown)


assets/js/chart.js
```diff
@@ -27,6 +28,22 @@ export function bindChartRangeLinks(accuracyForm) {
     chartRangeInput.value = 'Daily';
     accuracyForm.submit();
   });
+
+  var routeButtonElements = document.getElementsByClassName('route-button');
+  for (var i = 0; i < routeButtonElements.length; i++) {
+    routeButtonElements[i].addEventListener('click', function(event) {
+      event.preventDefault();
+      routeIdInput.value = this.text;
```
losvedir: I think `this` is brittle. It will refer to the object that actually received the click, not necessarily what you're expecting (relevant if, say, we add a `<span>` or `<em>` or something inside. I think better is to use `event.currentTarget` which always refers to the element to which the event listener is attached.

lemald: Fixed

assets/js/chart.js
```diff
@@ -27,6 +28,22 @@ export function bindChartRangeLinks(accuracyForm) {
     chartRangeInput.value = 'Daily';
     accuracyForm.submit();
   });
+
+  var routeButtonElements = document.getElementsByClassName('route-button');
+  for (var i = 0; i < routeButtonElements.length; i++) {
+    routeButtonElements[i].addEventListener('click', function(event) {
+      event.preventDefault();
+      routeIdInput.value = event.currentTarget.text;
```
losvedir: Hm, actually does this line even matter? It looks like `routeIdInput.value` gets set in both the `if` and `else` clauses below. (And those clauses also should be switched from `this`, btw.)

lemald: Haha, whooops. Fixed both issues.

PR #67, opened 2019-03-07
Change UI for selecting stop in prediction analyzer

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [[5] Analyzer: show stop name](https://app.asana.com/0/584764604969369/1110237990554878)

Changed the input for stop IDs from a text field to a combo box populated with the stop descriptions and stop IDs. 

#### Reviewer Checklist

### DIFF
--- a/assets/brunch-config.js
+++ b/assets/brunch-config.js
@@ -57,6 +57,10 @@ exports.config = {
   },
 
   npm: {
-    enabled: true
+    enabled: true,
+    globals: {
+      jQuery: "jquery/dist/jquery.min",
+      combobox: "\@danielfarrell/bootstrap-combobox/js/bootstrap-combobox"
+    }
   }
 };
--- a/config/config.exs
+++ b/config/config.exs
@@ -28,6 +28,7 @@ config :logger, :console,
 config :prediction_analyzer, ecto_repos: [PredictionAnalyzer.Repo]
 config :prediction_analyzer, http_fetcher: HTTPoison
 config :prediction_analyzer, :migration_task, Predictions.ReleaseTasks.NoOp
+config :prediction_analyzer, :stop_name_fetcher, PredictionAnalyzer.StopNameFetcher
 
 config :prediction_analyzer, PredictionAnalyzer.Repo,
   adapter: Ecto.Adapters.Postgres,
--- a/config/config.exs
+++ b/config/config.exs
@@ -28,6 +28,7 @@ config :logger, :console,
 config :prediction_analyzer, ecto_repos: [PredictionAnalyzer.Repo]
 config :prediction_analyzer, http_fetcher: HTTPoison
 config :prediction_analyzer, :api_v3_key, System.get_env("API_V3_KEY")
+config :prediction_analyzer, :stop_fetch_url, "https://api-v3.mbta.com/stops"
 config :prediction_analyzer, :migration_task, Predictions.ReleaseTasks.NoOp
 config :prediction_analyzer, :stop_name_fetcher, PredictionAnalyzer.StopNameFetcher
 
--- a/config/config.exs
+++ b/config/config.exs
@@ -27,6 +27,7 @@ config :logger, :console,
 
 config :prediction_analyzer, ecto_repos: [PredictionAnalyzer.Repo]
 config :prediction_analyzer, http_fetcher: HTTPoison
+config :prediction_analyzer, :api_v3_key, System.get_env("API_V3_KEY")
 config :prediction_analyzer, :migration_task, Predictions.ReleaseTasks.NoOp
 config :prediction_analyzer, :stop_name_fetcher, PredictionAnalyzer.StopNameFetcher
 
--- a/config/config.exs
+++ b/config/config.exs
@@ -27,7 +27,6 @@ config :logger, :console,
 
 config :prediction_analyzer, ecto_repos: [PredictionAnalyzer.Repo]
 config :prediction_analyzer, http_fetcher: HTTPoison
-config :prediction_analyzer, :api_v3_key, System.get_env("API_V3_KEY")
 config :prediction_analyzer, :stop_fetch_url, "https://api-v3.mbta.com/stops"
 config :prediction_analyzer, :migration_task, Predictions.ReleaseTasks.NoOp
 config :prediction_analyzer, :stop_name_fetcher, PredictionAnalyzer.StopNameFetcher
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -59,4 +59,4 @@ config :prediction_analyzer, PredictionAnalyzer.Repo,
   adapter: Ecto.Adapters.Postgres,
   pool_size: 10
 
-config :prediction_analyzer, start_workers: true
+config :prediction_analyzer, start_workers: false
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -59,4 +59,4 @@ config :prediction_analyzer, PredictionAnalyzer.Repo,
   adapter: Ecto.Adapters.Postgres,
   pool_size: 10
 
-config :prediction_analyzer, start_workers: false
+config :prediction_analyzer, start_workers: true
--- a/lib/prediction_analyzer/stop_name_fetcher.ex
+++ b/lib/prediction_analyzer/stop_name_fetcher.ex
@@ -2,21 +2,28 @@ defmodule PredictionAnalyzer.StopNameFetcher do
   require Logger
   use GenServer
 
+  @type state() :: [{String.t(), String.t()}]
+
+  @spec start_link([any]) :: {:ok, pid}
   def start_link(opts \\ []) do
     state = get_stop_names()
     GenServer.start_link(__MODULE__, state, opts)
   end
 
+  @spec init(state) :: {:ok, state}
   def init(state) do
     {:ok, state}
   end
 
+  @spec get_stop_map() :: {:reply, state, state}
   def get_stop_map() do
     GenServer.call(__MODULE__, :get_stop_map)
   end
 
+  @spec handle_call(:get_stop_map, GenServer.from(), state) :: {:reply, state, state}
   def handle_call(:get_stop_map, _from, state), do: {:reply, state, state}
 
+  @spec get_stop_names() :: state
   def get_stop_names do
     url = "https://api-v3.mbta.com/stops"
     api_key = Application.get_env(:prediction_analyzer, :api_v3_key)
@@ -35,6 +42,7 @@ defmodule PredictionAnalyzer.StopNameFetcher do
     end
   end
 
+  @spec parse_response(%HTTPoison.Response{}) :: state
   defp parse_response(http_response) do
     {:ok, body} = Jason.decode(http_response.body)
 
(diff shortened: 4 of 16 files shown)


config/config.exs
```diff
@@ -27,7 +27,10 @@ config :logger, :console,
 
 config :prediction_analyzer, ecto_repos: [PredictionAnalyzer.Repo]
 config :prediction_analyzer, http_fetcher: HTTPoison
+config :prediction_analyzer, :api_v3_key, System.get_env("API_V3_KEY")
```
losvedir: This won't work. The `System.get_env` calls in `config/` are done at _compile time_, which is Semaphore. To pick up the ENV vars on AWS, it needs to be done at run time. So we make these calls in `application.ex`'s `set_runtime_config` function.

assets/brunch-config.js
```diff
@@ -57,6 +57,10 @@ exports.config = {
   },
 
   npm: {
-    enabled: true
+    enabled: true,
+    globals: {
```
losvedir: Does it have to be global? Is there no way to `import` `jquery` and `combobox` in the relevant JS files?

phildarnowsky: Perhaps, I'll give it a try.

config/dev.exs
```diff
@@ -59,4 +59,4 @@ config :prediction_analyzer, PredictionAnalyzer.Repo,
   adapter: Ecto.Adapters.Postgres,
   pool_size: 10
 
-config :prediction_analyzer, start_workers: false
+config :prediction_analyzer, start_workers: true
```
losvedir: Is this intentional? I think I had put it as `false` so you don't have to specify a million environment variables at start up. But if people prefer to have all the workers going, then that's fine with me.

phildarnowsky: I think we did this for our convenience while developing and it can come out.

lib/prediction_analyzer/stop_name_fetcher.ex
```diff
@@ -0,0 +1,55 @@
+defmodule PredictionAnalyzer.StopNameFetcher do
+  require Logger
+  use GenServer
+
+  @type state() :: [{String.t(), String.t()}]
+
+  @spec start_link([any]) :: {:ok, pid}
+  def start_link(opts \\ []) do
+    state = get_stop_names()
```
losvedir: In general, we don't want to do too much work in the `start_link` or the `init` functions, since if it takes too long to reply with `{:ok, state}`, the supervisor and in turn the whole app start up will :boom:. We should only do logic that is absolutely necessary for the app starting up. I think here, we can afford to start with an empty state, and only make this API call after the app is up and running.

So can you move this `get_stop_names()` logic to a `handle_info(:get_stop_names, state)` GenServer callback? And then in `init` you can do `send self(), :get_stop_names`, which will send itself a message, which it will handle after it has finished starting up.

config/dev.exs
losvedir: Just FYI, if you did it because you needed data, you can run `mix ecto.reset` which will reset the DB schema and then seed it with sample data.

losvedir:
Running it locally, when I first load the page, the stop_id seems to be pre-filled as Airport. However, I think the data is actually the "all stops" data, and so the combobox is just prefilling the first element. Can you configure it to be blank if no stop_id is selected? Otherwise it's a little misleading.

phildarnowsky:
@losvedir yes I just noticed too that we need a blank option

assets/brunch-config.js
phildarnowsky: I only managed to get it working within `globals`.

lib/prediction_analyzer/stop_name_fetcher.ex
```diff
@@ -0,0 +1,55 @@
...
+    {:ok, state}
+  end
+
+  @spec get_stop_map() :: {:reply, state, state}
+  def get_stop_map(pid \\ __MODULE__) do
+    GenServer.call(pid, :get_stop_map)
+  end
+
+  @spec handle_call(:get_stop_map, GenServer.from(), state) :: {:reply, state, state}
+  def handle_call(:get_stop_map, _from, state), do: {:reply, state, state}
```
losvedir: It looks like the structure of the state here is something like:

```ex
[
  {"Airport - Blue Line - Bowdoin (70047)", "70047"},
  {"Airport - Blue Line - Wonderland (70048)", "70048"},
  {"Alewife - Red Line (70061)", "70061"},
  ...
]
```

That seems to me to be a little _too_ tailored to _this_ specific use case, and less useful for other uses down the road (in particular, the ticket we have on the sprint this week, about labeling the chart by stop).

The two issues I see are:

* It's a giant list, which works mostly if you're dumping all the stops, but means you can't look a given stop up effectively.

* The stop name "value" has been stored with the whole `name (id)` format, which seems to me to be a presentational concern that the frontend combobox or serialization step should be doing, and not the form that it should be stored in the GenServer.

I think it would make more sense and be more useful for the upcoming ticket to be stored in the form:

```ex
%{
  "70047" => "Airport - Blue Line - Bowdoin",
  "70048" => "Airport - Blue Line - Wonderland"
  # ...
}
```

What do you think?

phildarnowsky: @losvedir hmmm yes makes sense to me.

phildarnowsky: @losvedir the only thing that makes me a little leery of this idea is that we want the list to be sorted alphabetically, and by precalculating everything we only had to do so once, rather than on every request. But I think, even sorting on every request, the performance will be tolerable, and if not we'll do something a little more clever.

phildarnowsky:
@losvedir ready for another whirl

phildarnowsky:
@losvedir psyche, codecov hadn't finished yet, placating it now

phildarnowsky:
@losvedir good for real now.

lib/prediction_analyzer/stop_name_fetcher.ex
```diff
@@ -0,0 +1,70 @@
...
+
+  @spec init(state) :: {:ok, state}
+  def init(state) do
+    send(self(), :get_stop_names)
+    {:ok, state}
+  end
+
+  @spec get_stop_map() :: {:reply, state, state}
+  def get_stop_map(pid \\ __MODULE__) do
+    real_pid =
```
losvedir: This all shouldn't be necessary. You can just do `GenServer.call` with the atom, assuming you've started up the GenServer with a `name: ..` argument (which you're doing here).

lib/prediction_analyzer/stop_name_fetcher.ex
```diff
@@ -0,0 +1,70 @@
+defmodule PredictionAnalyzer.StopNameFetcher do
+  require Logger
+  use GenServer
+
+  @type state() :: %{String.t() => String.t()}
+
+  @spec start_link([any]) :: {:ok, pid}
+  def start_link(opts \\ []) do
+    GenServer.start_link(__MODULE__, [{"", ""}], opts)
```
losvedir: We usually define the initial state as part of the `init` callback, as it's generally considered an implementation detail which we shouldn't expose to callers. Also, I think the initial state should be `%{}` and not a tuple of empty strings?

lib/prediction_analyzer/stop_name_fetcher.ex
```diff
@@ -0,0 +1,70 @@
...
+
+    http_fetcher = Application.get_env(:prediction_analyzer, :http_fetcher)
+
+    case http_fetcher.get(url, headers, params: params) do
+      {:ok, response} ->
+        parse_response(response)
+
+      {:error, e} ->
+        Logger.warn("Could not download stop names; received: #{inspect(e)}")
+        %{"" => ""}
```
losvedir: Why do we want a fallback map of an empty string to an empty string? The "principle of least surprise" would be a `%{}`. Here, and in the `init` callback, and in the `Enum.into(%{})` call in `parse_response`.

lib/prediction_analyzer/stop_name_fetcher.ex
losvedir: Ah, I see. That's a good point, but I agree with your assessment that it's probably fine to do anyway and we can see how it goes.

lib/prediction_analyzer/stop_name_fetcher.ex
phildarnowsky: I meant for it to be a _map_ of an empty string to an empty string. The idea was to have, in the case that the fetch failed or similar, a fallback state that still allowed for a blank `<option>` tag. But there's no real advantage to having that since you bring it up.

lib/prediction_analyzer/stop_name_fetcher.ex
phildarnowsky: We do need to check `Process.alive?` to cover the situation where we start in dev mode and haven't set `start_workers` to `true` in `config/dev.exs`. Otherwise when the view calls this function, it tries to do a `GenServer.call` to a process that doesn't exist, and crashes.

However with a few tweaks some other of this stuff can come out.

lib/prediction_analyzer/stop_name_fetcher.ex
phildarnowsky: See comment above, I'll replace these with empty maps.

lib/prediction_analyzer/stop_name_fetcher.ex
phildarnowsky: Also I don't understand the part of your comment about not wanting to expose state to callers. How does setting the initial state in `start_link` expose that state to callers of `start_link`?

phildarnowsky:
@losvedir there we go

lib/prediction_analyzer/stop_name_fetcher.ex
```diff
@@ -0,0 +1,65 @@
...
+
+  @spec init(state) :: {:ok, state}
+  def init(state) do
+    send(self(), :get_stop_names)
+    {:ok, state}
+  end
+
+  @spec get_stop_map() :: {:reply, state, state}
+  def get_stop_map() do
+    pid = Process.whereis(__MODULE__)
```
losvedir: Oh, I meant the entirely of this function should just be `GenServer.call(pid, :get_stop_map)`. See, e.g., our RTR [Vehicles](https://github.com/mbta/rtr/blob/master/rtr/lib/rtr/vehicles/state.ex#L36), [Alerts](https://github.com/mbta/rtr/blob/master/rtr/lib/rtr/alerts/state.ex#L51), or [Segments](https://github.com/mbta/rtr/blob/master/rtr/lib/rtr/segments/state.ex#L27) GenServers.

The "client" functions of a GenServer should be nothing more than a pass through to invoke the "server" callbacks.

phildarnowsky: Sure, makes sense, but as I explained above in dev mode this GenServer will not actually be running, and if we try to `GenServer.call(pid, :get_stop_map)` in that case, the application will crash.

The other alternative I see is to check that the server is running before calling `get_stop_map` but that seems even less appealing.

losvedir:
Oohh, I see. You changed the dev environment's `:start_workers` to true because the stop API state is a worker. Yeah, in contrast to the other ones, this one is probably useful to have running even in dev, so I think it makes sense to put it back how you had it. Sorry about that!

PR #68, opened 2019-03-07
Translate terminal stop IDs

lemald:
#### Summary of changes
**Asana Ticket:** [analyzer: handle new child stops](https://app.asana.com/0/584764604969369/1112207717058010/f)

Added a new utility function to handle the mapping itself. The stop IDs now get translated when downloading either the predictions or vehicle positions JSON, so all of the data stored in prediction analyzer will be in terms of the more generic ID.

#### Reviewer Checklist

### DIFF
--- a/lib/fake_httpoison.ex
+++ b/lib/fake_httpoison.ex
@@ -28,7 +28,7 @@ defmodule FakeHTTPoison do
             },
             "trip" => %{
               "trip_id" => "TEST_TRIP",
-              "route_id" => "Blue"
+              "route_id" => "Red"
             },
             "stop_time_update" => [
               %{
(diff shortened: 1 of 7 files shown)


lib/fake_httpoison.ex
```diff
@@ -28,7 +28,7 @@ defmodule FakeHTTPoison do
             },
             "trip" => %{
               "trip_id" => "TEST_TRIP",
```
dmaltzan: Just curious - which test/tests are affected by this change?

lemald: As far as I'm aware it's just the `PredictionAnalyzer.Predictions.Download` tests for `get_predictions` that use the specified HTTP fetcher.

PR #72, opened 2019-03-11
Handle simultaneous departures / arrivals

lemald:
#### Summary of changes
**Asana Ticket:** [Investigate Bug: Why do Red Line Park Street NB and DTX NB have low accuracy?](https://app.asana.com/0/584764604969369/1110212001889553/f)

In practice, we don't get location data for Red Line trains at any point between the platforms at Downtown Crossing and Park Street, meaning that they transition from stopped at DTX to stopped at Park with nothing in between. We need to add a new clause to `compare_vehicle` to handle this case and record arrival / departure times appropriately.

#### Reviewer Checklist

### DIFF
--- a/lib/prediction_analyzer/vehicle_positions/comparator.ex
+++ b/lib/prediction_analyzer/vehicle_positions/comparator.ex
@@ -43,7 +43,7 @@ defmodule PredictionAnalyzer.VehiclePositions.Comparator do
          %Vehicle{stop_id: old_stop, current_status: old_status} = old_vehicle
        )
        when new_stop != old_stop and old_status == :STOPPED_AT and new_status == :STOPPED_AT do
-    record_departure(old_vehicle)
+    record_departure(%{old_vehicle | timestamp: new_vehicle.timestamp})
     record_arrival(new_vehicle)
   end
 
--- a/lib/prediction_analyzer/vehicle_positions/comparator.ex
+++ b/lib/prediction_analyzer/vehicle_positions/comparator.ex
@@ -31,11 +31,11 @@ defmodule PredictionAnalyzer.VehiclePositions.Comparator do
   end
 
   defp compare_vehicle(
-         %Vehicle{stop_id: new_stop, current_status: new_status},
+         %Vehicle{stop_id: new_stop, current_status: new_status, timestamp: new_timestamp},
          %Vehicle{stop_id: old_stop, current_status: old_status} = old_vehicle
        )
        when new_stop != old_stop and old_status == :STOPPED_AT and new_status != :STOPPED_AT do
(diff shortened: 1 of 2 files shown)


lib/prediction_analyzer/vehicle_positions/comparator.ex
```diff
@@ -37,6 +38,15 @@ defmodule PredictionAnalyzer.VehiclePositions.Comparator do
     record_departure(old_vehicle)
   end
 
+  defp compare_vehicle(
+         %Vehicle{stop_id: new_stop, current_status: new_status} = new_vehicle,
+         %Vehicle{stop_id: old_stop, current_status: old_status} = old_vehicle
+       )
+       when new_stop != old_stop and old_status == :STOPPED_AT and new_status == :STOPPED_AT do
+    record_departure(old_vehicle)
```
losvedir: Should this be based on `old_vehicle` or `new_vehicle`? I think by using `old_vehicle` here, we're basically saying that the vehicle arrived at and departed from the previous stop at the same time. Whereas if we use `new_vehicle` we're saying that it departed from the previous stop at the same time that it arrived at the next stop, which seems more accurate to me.

lemald: I went back and forth on this a couple of times, but keep in mind that `old_vehicle` is the last update we received from the old stop ID before it changed over to the new stop ID, so it's actually recording the departure from the first stop as the last moment it was recorded as being on the platform there. This is actually analogous to how the other call to `record_departure` does it. The expected output in this case would be a departure from stop 1 followed maybe about a second later by an arrival at stop 2.

losvedir: Hmm, I'm not sure about that. It updates based on the vehicle's timestamp, which I believe is set in the VehiclePositions file based on the timestamp in the last TMOV we received. So if that vehicle basically went TMOV stop1 -> TMOV stop2, then isn't the timestamp in the first vehicle, the original one that that vehicle had when it arrived at stop1?

I still think that the options are as follows:

* Use `old_vehicle` here: the arrival_time and departure_time for stop1 are the same, and the arrival_time at stop2 is now-ish.
* Use `new_vehicle` here: the arrival_time at stop1 is old, and the departure_time from stop1 and the arrival_time at stop2 are the same and now-ish.

lemald: Ah, okay, I just didn't realize exactly how the timestamp in question worked, so I'll change the logic. However, it looks like this means the existing departure time logic is flawed. Sure enough, I've looked at some random examples and every case I've seen, the departure time is the same as the arrival time. Do you think I should update the other case as well?

PR #73, opened 2019-03-11
Sort bin options in proper order

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [[1] Sort bins correctly in prediction analyzer](https://app.asana.com/0/584764604969369/1112192658166930)

Bins were appearing in an arbitrary order in the dropdown for same, make sure they're in the natural order.

#### Reviewer Checklist

### DIFF
--- a/lib/prediction_analyzer_web/views/accuracy_view.ex
+++ b/lib/prediction_analyzer_web/views/accuracy_view.ex
@@ -11,6 +11,17 @@ defmodule PredictionAnalyzerWeb.AccuracyView do
     0
   end
 
+  @spec bin_options() :: [String.t()]
+  def bin_options() do
+    PredictionAccuracy.bins()
+    |> Map.keys()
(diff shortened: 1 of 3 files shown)


lib/prediction_analyzer_web/views/accuracy_view.ex
```diff
@@ -11,6 +11,17 @@ defmodule PredictionAnalyzerWeb.AccuracyView do
     0
   end
 
+  @spec bin_options() :: [String.t()]
+  def bin_options() do
```
lemald: You should probably write a quick-and-dirty test for this - nothing too fancy, but since it is business logic that we're adding it should probably be tested somehow.

phildarnowsky: It'll take about 5 seconds to write a test for this, better to have it than not.

phildarnowsky:
@lemald OK added the test

PR #74, opened 2019-03-13
Accuracy-by-stop view

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [[5] analyzer: new view comparing stops on a route](https://app.asana.com/0/584764604969369/950496618789686)

A third view is added to the prediction analyzer, where it breaks down accuracy on a stop-by-stop basis.

#### Reviewer Checklist

### DIFF
--- a/assets/js/chart.js
+++ b/assets/js/chart.js
@@ -52,28 +52,36 @@ export function setupDashboard() {
   var dgAccs = rawData["dg_accs"];
   var dateRangeData = rawData["buckets"];
   var chartType = rawData["chart_type"];
+  var chartHeight;
   var dataType;
+  var rotateAxes;
   var xAxisText;
   var xAxisType;
   var xAxisRotation;
 
   switch(chartType) {
     case "Hourly": {
+      chartHeight = 540;
       dataType = "line";
+      rotateAxes = false;
       xAxisText = "Hour of Day";
       xAxisType = "indexed";
       xAxisRotation = 0;
       break;
     }
     case "Daily": {
+      chartHeight = 540;
       dataType = "line";
+      rotateAxes = false;
       xAxisText = "";
       xAxisType = "timeseries";
       xAxisRotation = 75;
       break;
     }
     case "By Station": {
+      chartHeight = prodAccs * 25;
       dataType = "bar";
+      rotateAxes = true;
       xAxisText = "";
       xAxisType = "category";
       xAxisRotation = 90;
@@ -100,6 +108,7 @@ export function setupDashboard() {
       pattern: ["#1fecff", "#c743f0"]
     },
     axis: {
+      rotated: rotateAxes,
       y: {
         label: {
           text: "% Accurate",
@@ -135,7 +144,7 @@ export function setupDashboard() {
       }
     },
     size: {
-      height: 540
+      height: chartHeight
     },
     legend: {
       position: "inset"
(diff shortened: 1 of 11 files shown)


losvedir:
I pulled this down and it looks pretty rad with the sample data! Have you tried it with a more realistic number of stations? Does it scroll? Get squashed?

Screen Shot 2019-03-13 at 2 29 20 PM

phildarnowsky:
@losvedir I tried it with the B branch, which I believe has the most stops of any line, and got this, which is crowded but still legible. Notice how some of the names are cut off, but I think it's acceptable.
Screen Shot 2019-03-13 at 4 14 30 PM

Now on the other hand, this is what you get if you do _all_ lines, probably we'll want to...not do that. I'm not sure of what's best in this case, whether we might want to leave off the legend, or just forbid using this view altogether unless you have a line selected.
Screen Shot 2019-03-13 at 4 16 04 PM

losvedir:
Perhaps I'm missing it in the code, but I don't believe it does this aspect of the ticket?

>  The chart has on its axis every stop, with its name (not child stop ID) **but ordered by child stop ID**, and the chart displays as a bar chart.

In lieu of building out the actual patterns of the routes, we said that ordering by the stop ID would be a quick and dirty approximation to ensure that the stops are roughly in order.

losvedir:
Ha, wow! That really is a lot of stops... just out of curiosity (and I know Dave will hate this...) what does it look like if you [rotate the axes](https://c3js.org/samples/axes_rotated.html)?

phildarnowsky:
I think too that the all-stops view would be useful, and I was thinking, what if we just didn't show the labels in that case? The tooltips still work so you still have some way of telling which bar is associated with which stop.

phildarnowsky:
@losvedir no I don't believe I am sorting by ID, good catch

phildarnowsky:
@losvedir I tried the chart `rotated` and even so, I have to make the chart five times taller before the labels are legible, which I don't think we want to do. So I'm inclined to pursue another solution, what do you think of the idea of leaving the labels off if there are more than N?

phildarnowsky:
@losvedir wait, I lied, it is sorting by child ID, pre-existing code already does that. `lib/prediction_analyzer/prediction_accuracy/prediction_accuracy.ex`, lines 150 through 161.

losvedir:
@phildarnowsky - okay, one last idea, and then we can go with just not having labels. I was able to make it work and look pretty good, with rotated axes and `size: { height: prodAccs.length * 25 } }`. What do you think of that solution? I just hamfistedly tossed it in there to see how it would look, but I guess you'd have to change the configuration a little bit to only do this in the "By Station" mode.

phildarnowsky:
@losvedir I like it! Let's go with that.

phildarnowsky:
@losvedir OK I implemented your suggestion, anything further on this branch?

assets/js/chart.js
```diff
@@ -49,8 +50,44 @@ export function setupDashboard() {
...
+      chartHeight = 540;
+      dataType = "line";
+      rotateAxes = false;
+      xAxisText = "";
+      xAxisType = "timeseries";
+      xAxisRotation = 75;
+      break;
+    }
+    case "By Station": {
+      chartHeight = prodAccs * 25;
```
losvedir: I think this should be `prodAccs.length`. That compresses the chart a little more to my liking. But if you're sold on the height that it is, you can adjust `25`. Either way it should be `.length`.

phildarnowsky: whoops, fixed

PR #76, opened 2019-03-18
Add direction ID

lemald:
#### Summary of changes
**Asana Ticket:** [Prediction analyzer filter by direction](https://app.asana.com/0/584764604969369/1002946782363492/f)

Added direction ID to the downloading, processing, and aggregating. Put a new field on the form and support filtering.

#### Reviewer Checklist


lib/prediction_analyzer_web/templates/accuracy/index.html.eex
```diff
@@ -72,6 +72,22 @@
           <% end %>
         </div>
 
+	<div class="form-group">
+	    <%= label(f, :direction_id, "Direction ID") %>
```
losvedir: Spacing here is wonky.

Regarding your question about how it looks, well I think calling this a "label" is not quite correct. Labels are associated with input elements, and so the ones below for `"0"`, `"1"` and `"Any"` are fine. I think the "right" thing here to group the radios in a fieldset and give it a legend or something, but in lieu of that, I think it's probably fine to just have `<div><strong>Direction ID</strong></div>`.

lemald: Hah, right, I was sort of on auto-pilot when copying the general layout of the other parts of the form. That approach works fine (and I fixed the indentation). Thanks!

lib/prediction_analyzer/prediction_accuracy/prediction_accuracy.ex
```diff
@@ -94,6 +97,14 @@ defmodule PredictionAnalyzer.PredictionAccuracy.PredictionAccuracy do
 
   defp filter_by_arrival_departure(q, _), do: {:ok, q}
 
+  @spec filter_by_direction(Ecto.Query.t(), any()) :: {:ok, Ecto.Query.t()} | {:error, String.t()}
+  defp filter_by_direction(q, direction_id) when direction_id in ["0", "1"] do
```
sghiaa: i realize this is a pattern we use for every field but it occurs to me that its kind of weird that we rely on the fact that "any" from the form will not match this array, and therefore be picked up by the other definition of this function.  i wonder if it should be made explicit?  leaving the base case function too, of course so its really just extra work but it makes the relationship clearer maybe.

sghiaa: im ok if we do nothing, though

lib/prediction_analyzer/prediction_accuracy/query.ex
```diff
@@ -136,6 +137,7 @@ defmodule PredictionAnalyzer.PredictionAccuracy.Query do
         $2 AS hour_of_day,
         p.route_id AS route_id,
         p.stop_id AS stop_id,
+	p.direction_id AS direction_id,
```
sghiaa: this seems to be 1 space too far to theleft, as does the direction id above

lemald: Should be fixed.

lib/prediction_analyzer/prediction_accuracy/prediction_accuracy.ex
lemald: Hmm, this is a good point, though I'm not sure that making it more explicit is worth adding an additional function clause if it's just going to be a copy of the existing one. I'm probably going to leave it for now but I'll consider how we could do it better in the future.

lib/prediction_analyzer/prediction_accuracy/query.ex
```diff
@@ -125,6 +125,7 @@ defmodule PredictionAnalyzer.PredictionAccuracy.Query do
         hour_of_day,
         route_id,
         stop_id,
+        direction_id,
```
sghiaa: i think technically by rule this hsould get a typespec.

lemald: Good catch, added.

sghiaa:
it looks like we dont have a test that specifically checks for direction id. if we add that, i think this ✔️ 's all the 🔲 's

PR #79, opened 2019-03-25
Record arrival for train that materializes STOPPED_AT a terminal

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [[2] analyzer: handle first trip of the day](https://app.asana.com/0/584764604969369/1113885023269820)

There are a great many trains in the early hours that appear to not have their arrival at the terminal tracked, so from prediction analyzer's point of view they seem to just appear at the terminal out of nowhere, with a STOPPED_AT status. This results in their departures not being correctly recorded, and our prediction accuracy suffers.

#### Reviewer Checklist

### DIFF
--- a/lib/prediction_analyzer/vehicle_events/vehicle_event.ex
+++ b/lib/prediction_analyzer/vehicle_events/vehicle_event.ex
@@ -29,6 +29,7 @@ defmodule PredictionAnalyzer.VehicleEvents.VehicleEvent do
     field(:departure_time, :integer)
   end
 
+  @spec changeset(Ecto.Schema.t(), %{atom => any}) :: Ecto.Changeset.t()
   def changeset(vehicle_event, params \\ %{}) do
     fields = [
       :vehicle_id,
--- a/lib/prediction_analyzer/vehicle_events/vehicle_event.ex
+++ b/lib/prediction_analyzer/vehicle_events/vehicle_event.ex
@@ -42,8 +42,10 @@ defmodule PredictionAnalyzer.VehicleEvents.VehicleEvent do
       :arrival_time
     ]
 
+    required_fields = fields |> Enum.reject(&(&1 == :arrival_time))
+
     vehicle_event
     |> cast(params, fields)
-    |> validate_required(fields)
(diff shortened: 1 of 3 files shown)


lib/prediction_analyzer/vehicle_positions/comparator.ex
```diff
@@ -47,7 +47,11 @@ defmodule PredictionAnalyzer.VehiclePositions.Comparator do
     record_arrival(new_vehicle)
   end
 
-  defp compare_vehicle(%Vehicle{label: label}, nil) do
+  defp compare_vehicle(%Vehicle{label: label, current_status: status} = vehicle, nil) do
+    if status == :STOPPED_AT do
+      record_arrival(vehicle)
```
lemald: I'm a little bit concerned about using `record_arrival` as-is since I don't think it makes sense to save an arrival time for these trains. I think it would be best to have a new function called something like `record_new_stopped_vehicle` that creates a vehicle event without an arrival time, then test that the appropriate combination of calls to `compare_vehicle` for its initial position and then its departure results in a vehicle event with no arrival time but a departure time.

lemald: Note that this approach will also require tweaking `record_departure` slightly.

lemald:
Looks generally good to me, but not to codecov. If you fix the codecov thing I should be pretty much ready to approve it though.

phildarnowsky:
@lemald codecov is happy, life is good

lib/prediction_analyzer/vehicle_events/vehicle_event.ex
```diff
@@ -42,8 +42,10 @@ defmodule PredictionAnalyzer.VehicleEvents.VehicleEvent do
       :arrival_time
     ]
 
+    required_fields = fields |> Enum.reject(&(&1 == :arrival_time))
+
```
lemald: Sorry, one last thing - I think this means `changeset` needs a typespec, since it doesn't seem to be an implementation of a callback or anything.

PR #82, opened 2019-03-28
Sort by-station view by accuracy or ID

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [[3] Prediction Analyzer sort by accuracy](https://app.asana.com/0/584764604969369/1114524503205260)

Added a button-style link to the "By Station" view that sorts the chart by either accuracy or route ID, without the need for a page refresh.

#### Reviewer Checklist

### DIFF
--- a/assets/js/chart.js
+++ b/assets/js/chart.js
@@ -64,12 +64,11 @@ function renderDashboard() {
 
   sortedDataPoints = window.dataPoints.sort(sortFunction);
 
-  for(i = 0; i < sortedDataPoints.length; i++) {
-    var dataPoint = sortedDataPoints[i];
+  sortedDataPoints.forEach(function(dataPoint) {
     sortedProdAccs.push(dataPoint.prodAcc);
     sortedDgAccs.push(dataPoint.dgAcc);
     sortedBucketNames.push(dataPoint.bucket);
-  }
+  });
 
   switch(window.chartType) {
     case "Hourly": {
--- a/assets/js/chart.js
+++ b/assets/js/chart.js
@@ -40,7 +40,6 @@ function renderDashboard() {
   var xAxisText;
   var xAxisType;
   var xAxisRotation;
-  var sortedDataPoints;
   var sortFunction;
   var sortOrderLink;
   var sortOrder;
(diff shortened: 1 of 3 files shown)


assets/js/chart.js
```diff
@@ -30,20 +30,48 @@ export function bindFormLinks(accuracyForm) {
   bindChartRangeLink('link-by_station', 'By Station');
 }
 
-export function setupDashboard() {
-  var rawData = window.dataPredictionAccuracyJSON;
-  var prodAccs = rawData["prod_accs"];
-  var dgAccs = rawData["dg_accs"];
-  var dateRangeData = rawData["buckets"];
-  var chartType = rawData["chart_type"];
+export function renderDashboard() {
```
losvedir: Do this function and `toggleSortOrder` need to be exported? It seems like they're only called from within this file, and only `setupDashboard` is called elsewhere. Or is this a brunch requirement? I forget how it exports stuff compared to the usual module exports of ES6 from webpack.

phildarnowsky: No, works fine without the `export`s.

phildarnowsky:
@losvedir do you have any other feedback on this PR?

assets/js/chart.js
```diff
@@ -30,20 +30,48 @@ export function bindFormLinks(accuracyForm) {
...
+    sortOrder = "by_id";
+  }
+
+  if(sortOrder == "by_id") {
+    sortFunction = function(a, b) { return (a.id < b.id) ? -1 : 1; }
+  } else {
+    sortFunction = function(a, b) { return ([a.prodAcc, a.dgAcc] < [b.prodAcc, b.dgAcc]) ? -1 : 1; }
+  }
+
+  sortedDataPoints = window.dataPoints.sort(sortFunction);
```
losvedir: I think this assignment is misleading since the `sort` method actually mutates the underlying array. Maybe it would be better without this variable at all? Just `window.dataPoints.sort(sortFunction)` and then basing all the other stuff off `window.dataPoints`.

assets/js/chart.js
```diff
@@ -30,20 +30,48 @@ export function bindFormLinks(accuracyForm) {
...
+
+  for(i = 0; i < sortedDataPoints.length; i++) {
```
losvedir: I think this can use `forEach`. It's supported by all modern browsers, at least the ones I can see using this dashboard.

PR #89, opened 2019-04-22
Use If-Modified-Since header when querying CR positions

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [[2] analyzer: Send if-modified-since headers when querying API](https://app.asana.com/0/584764604969369/1118061165588451)

In an attempt to reduce the number of "One departure, multiple update" warnings when getting CR positions, we make use of the If-Modified-Since header when requesting CR positions from the API, to avoid overwriting more recent state with less recent.

#### Reviewer Checklist

### DIFF
--- a/lib/prediction_analyzer/vehicle_positions/tracker.ex
+++ b/lib/prediction_analyzer/vehicle_positions/tracker.ex
@@ -193,7 +193,7 @@ defmodule PredictionAnalyzer.VehiclePositions.Tracker do
     Process.send_after(pid, :track_commuter_rail_vehicles, 10_500)
   end
 
-  @spec extract_last_modified([tuple()]) :: String.t() | nil
+  @spec extract_last_modified([{String.t(), String.t()}]) :: String.t() | nil
   defp extract_last_modified(headers) do
     case Enum.find(headers, fn {key, _value} -> String.downcase(key) == "last-modified" end) do
       {_key, value} -> value
--- a/lib/prediction_analyzer/vehicle_positions/tracker.ex
+++ b/lib/prediction_analyzer/vehicle_positions/tracker.ex
@@ -42,7 +42,8 @@ defmodule PredictionAnalyzer.VehiclePositions.Tracker do
   def init(args) do
     state = Map.merge(args, %{subway_vehicles: %{}, commuter_rail_vehicles: %{}})
 
-    schedule_fetch(self())
+    schedule_subway_fetch(self())
+    schedule_commuter_rail_fetch(self())
     {:ok, state}
   end
 
@@ -61,7 +62,7 @@ defmodule PredictionAnalyzer.VehiclePositions.Tracker do
       end)
 
     Logger.info("Processed #{length(Map.keys(new_vehicles))} vehicles in #{time / 1000} ms")
-    schedule_fetch(self())
+    schedule_subway_fetch(self())
     {:noreply, %{state | subway_vehicles: new_vehicles}}
   end
 
@@ -76,6 +77,8 @@ defmodule PredictionAnalyzer.VehiclePositions.Tracker do
       "filter[route]" =>
         :commuter_rail |> PredictionAnalyzer.Utilities.routes_for_mode() |> Enum.join(",")
     }
+
+    schedule_commuter_rail_fetch(self())
 
     state =
       case PredictionAnalyzer.Utilities.APIv3.request(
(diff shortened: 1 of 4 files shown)


lib/prediction_analyzer/vehicle_positions/tracker.ex
```diff
@@ -12,7 +12,8 @@ defmodule PredictionAnalyzer.VehiclePositions.Tracker do
           environment: String.t(),
           aws_vehicle_positions_url: String.t(),
           subway_vehicles: vehicle_map(),
-          commuter_rail_vehicles: vehicle_map()
+          commuter_rail_vehicles: vehicle_map(),
+          commuter_rail_last_modified: String.t()
```
dmaltzan: Why only `commuter_rail_last_modified`? Shouldn't we also add `subway_last_modified` and log if this happens on subway? I wouldn't be surprised if it happened occasionally

phildarnowsky: @dmaltzan OK done.

phildarnowsky: @dmaltzan ...apart from the fact that now Codecov has to stick its oar in.

dmaltzan: @lemald can you do this one for me, thanks

dmaltzan: i mean finish this review

lemald: @dmaltzan yup, I was just starting to look at it.

lib/prediction_analyzer/vehicle_positions/tracker.ex
```diff
@@ -12,7 +12,9 @@ defmodule PredictionAnalyzer.VehiclePositions.Tracker do
           environment: String.t(),
           aws_vehicle_positions_url: String.t(),
           subway_vehicles: vehicle_map(),
-          commuter_rail_vehicles: vehicle_map()
+          commuter_rail_vehicles: vehicle_map(),
+          commuter_rail_last_modified: String.t(),
+          subway_last_modified: String.t()
```
lemald: Minor style nitpick, but this could go directly after `subway_vehicles`, or immediately above `commuter_rail_last_modified`, something , anything to make the ordering of the fields consistent. :-P

lib/prediction_analyzer/vehicle_positions/tracker.ex
```diff
@@ -131,8 +183,21 @@ defmodule PredictionAnalyzer.VehiclePositions.Tracker do
...
     Process.send_after(pid, :track_subway_vehicles, 1_000)
-    Process.send_after(pid, :track_commuter_rail_vehicles, 1_000)
+  end
+
+  @spec schedule_commuter_rail_fetch(pid() | atom()) :: reference
+  defp schedule_commuter_rail_fetch(pid) do
+    Process.send_after(pid, :track_commuter_rail_vehicles, 10_500)
+  end
+
+  @spec extract_last_modified([tuple()]) :: String.t() | nil
```
lemald: I think the `tuple()` type could be made more specific... isn't it always going to be of the form `{String.t(), String.t()}` or something similar?

test/prediction_analyzer/vehicle_positions/tracker_test.exs
```diff
@@ -38,23 +39,53 @@ defmodule PredictionAnalyzer.VehiclePositions.TrackerTest do
         aws_vehicle_positions_url: "vehiclepositions",
         environment: "dev-green",
         subway_vehicles: %{},
-        commuter_rail_vehicles: %{}
+        commuter_rail_vehicles: %{},
+        subway_last_modified: "Thu, 01 Jan 1970 00:00:00 GMT"
```
lemald: No `commuter_rail_last_modified` here?

lemald: Oh, I guess this is specifically for `track_subway_vehicles` so it doesn't matter.

phildarnowsky: It's not necessary for this test...probably neither is `commuter_rail_vehicles` now that I look at it

lemald: Yeah, agreed, you should be able to get rid of it. Alex and I must have been over-zealous with adding that field in there.

lib/utilities/api_v3.ex
```diff
@@ -1,10 +1,13 @@
 defmodule PredictionAnalyzer.Utilities.APIv3 do
   @default_opts [timeout: 2000, recv_timeout: 2000]
 
-  @spec request(String.t(), Keyword.t()) :: {:error, any()} | {:ok, map()}
-  def request(path, opts) do
+  @spec request(String.t(), [tuple()], Keyword.t()) :: {:error, any()} | {:ok, map()}
```
lemald: Oh, sorry for only just catching this, but I think the same basic comment about the `tuple()` type applies here as well.

PR #90, opened 2019-04-23
Remove empty parentheses from CR stop names

dmaltzan:
#### Summary of changes
**Asana Ticket:** [analyzer: why do commuter-rail stations have empty parentheses at the end?](https://app.asana.com/0/107495983514632/1117239379853577)

Handled case where `stop.platform_name` is nil

#### Reviewer Checklist

### DIFF
--- a/lib/fake_httpoison.ex
+++ b/lib/fake_httpoison.ex
@@ -48,7 +48,16 @@ defmodule FakeHTTPoison do
         params: %{"filter[route_type]" => "2"}
       ) do
     body = %{
-      "data" => []
+      "data" => [
+        %{
+          "id" => "Andover",
+          "attributes" => %{
+            "description" => nil,
+            "name" => "Andover",
+            "platform_name" => nil
+          }
+        }
+      ]
     }
 
     {:ok, %HTTPoison.Response{status_code: 200, body: Jason.encode!(body)}}
--- a/lib/fake_httpoison.ex
+++ b/lib/fake_httpoison.ex
@@ -56,14 +56,6 @@ defmodule FakeHTTPoison do
             "name" => "Andover",
             "platform_name" => nil
           }
-        },
-        %{
-          "id" => "99999999",
-          "attributes" => %{
-            "description" => nil,
-            "name" => nil,
-            "platform_name" => nil
-          }
         }
       ]
     }
--- a/lib/fake_httpoison.ex
+++ b/lib/fake_httpoison.ex
@@ -56,6 +56,14 @@ defmodule FakeHTTPoison do
             "name" => "Andover",
             "platform_name" => nil
           }
+        },
+        %{
+          "id" => "99999999",
+          "attributes" => %{
+            "description" => nil,
+            "name" => nil,
+            "platform_name" => nil
+          }
         }
       ]
     }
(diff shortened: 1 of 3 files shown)


lib/prediction_analyzer/stop_name_fetcher.ex
```diff
@@ -37,7 +37,20 @@ defmodule PredictionAnalyzer.StopNameFetcher do
 
   def handle_call({:get_stop_name, mode, stop_id}, _from, state) do
     stop = state[mode][stop_id]
-    stop_name = if stop, do: "#{stop.name} (#{stop.platform_name})", else: stop_id
+
+    IO.puts("stop_id = #{inspect(stop_id)} and stop = #{inspect(stop)}")
```
sghiaa: this seems like debug code

lib/prediction_analyzer/stop_name_fetcher.ex
```diff
@@ -37,7 +37,20 @@ defmodule PredictionAnalyzer.StopNameFetcher do
...
+
+    stop_name =
+      if stop do
```
sghiaa: im not a fan of nested if's; this could be written as a case statement

lib/prediction_analyzer/stop_name_fetcher.ex
```diff
@@ -36,8 +36,18 @@ defmodule PredictionAnalyzer.StopNameFetcher do
   end
 
   def handle_call({:get_stop_name, mode, stop_id}, _from, state) do
-    stop = state[mode][stop_id]
-    stop_name = if stop, do: "#{stop.name} (#{stop.platform_name})", else: stop_id
+    stop_name =
+      case state[mode][stop_id] do
```
sghiaa: if the mode is not in :commuter_rail and :subway, this will break.

dmaltzan: I mean, that was true before my PR too, but I'll add something

sghiaa: currently the only place this is called can only be with subway/commuter_rail but this is safer

lib/fake_httpoison.ex
```diff
@@ -48,7 +48,16 @@ defmodule FakeHTTPoison do
         params: %{"filter[route_type]" => "2"}
       ) do
     body = %{
-      "data" => []
+      "data" => [
```
sghiaa: this function doesnt have a typespec

lib/prediction_analyzer/stop_name_fetcher.ex
```diff
@@ -35,12 +35,27 @@ defmodule PredictionAnalyzer.StopNameFetcher do
     {:reply, description_map, state}
   end
 
-  def handle_call({:get_stop_name, mode, stop_id}, _from, state) do
-    stop = state[mode][stop_id]
-    stop_name = if stop, do: "#{stop.name} (#{stop.platform_name})", else: stop_id
+  def handle_call({:get_stop_name, mode, stop_id}, _from, state)
```
sghiaa: this function also doesnt have a typespec, but i cant remember if we normally spec our handle_x's

PR #91, opened 2019-04-24
Display stop ID when other fields are ambiguous

lemald:
#### Summary of changes
**Asana Ticket:** [Fix issue where some stops in prediction analyzer "by station" view have more than 2 bars associated](https://app.asana.com/0/584764604969369/1116298449952961/f)

It turns out there isn't an easy fix for the underlying GTFS data, so let's disambiguate with the stop ID when the other fields are all the same.

#### Reviewer Checklist

### DIFF
--- a/lib/prediction_analyzer/stop_name_fetcher.ex
+++ b/lib/prediction_analyzer/stop_name_fetcher.ex
@@ -97,7 +97,7 @@ defmodule PredictionAnalyzer.StopNameFetcher do
   defp parse_response(http_response) do
     {:ok, body} = Jason.decode(http_response.body)
 
-    unique_info =
+    unique_info_map =
       body["data"]
       |> Enum.reduce(%{}, fn stop, acc ->
         Map.update(
@@ -120,7 +120,7 @@ defmodule PredictionAnalyzer.StopNameFetcher do
         platform_name: stop["attributes"]["platform_name"]
       }
 
-      {stop["id"], Map.merge(stop_info, %{unique_info?: unique_info[stop_info]})}
+      {stop["id"], Map.merge(stop_info, %{unique_info?: unique_info_map[stop_info]})}
     end)
     |> Enum.into(%{})
   end
--- a/lib/prediction_analyzer/stop_name_fetcher.ex
+++ b/lib/prediction_analyzer/stop_name_fetcher.ex
@@ -2,7 +2,12 @@ defmodule PredictionAnalyzer.StopNameFetcher do
   require Logger
   use GenServer
 
-  @type state() :: %{PredictionAnalyzer.Utilities.mode() => %{String.t() => String.t()}}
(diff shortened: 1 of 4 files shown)


lib/prediction_analyzer_web/controllers/accuracy_controller.ex
```diff
@@ -148,6 +148,7 @@ defmodule PredictionAnalyzerWeb.AccuracyController do
       |> Map.put(:dg_accs, acc[:dg_accs] ++ dg_accuracy)
     end)
     |> Map.put(:chart_type, filter_params["chart_range"] || "Hourly")
+    |> IO.inspect(label: :chart_data)
```
phildarnowsky: @lemald presumably-stray `IO.inspect` here

lib/prediction_analyzer/stop_name_fetcher.ex
```diff
@@ -94,4 +107,23 @@ defmodule PredictionAnalyzer.StopNameFetcher do
...
+  @spec unique_stop_info?(%{String.t() => stop_info()}, stop_info()) :: boolean()
+  defp unique_stop_info?(mode_state, stop_info) do
+    mode_state
+    |> Map.values()
+    |> do_unique_stop_id?(stop_info, false)
+  end
+
+  @spec do_unique_stop_id?([stop_info()], stop_info(), boolean()) :: boolean()
+  defp do_unique_stop_id?([], _, _), do: true
+  defp do_unique_stop_id?([stop_info | _], stop_info, true), do: false
```
phildarnowsky: @lemald this is a somewhat inefficient implementation, compounded by the fact that we might call `get_stop_name` a few hundred times in a row in the worst case. What if instead, at the time that we download and parse the data, we build a map where the keys are the stuff that goes into `stop_info` and the values are the number of times that `stop_info` is seen in the data? Then we can retain that map as part of the state, and if we want to know if a given `stop_info` is unique, we just see if the corresponding value in that map is `1`.

phildarnowsky: Oh and shouldn't it be `do_unique_stop_info?` rather than `do_unique_stop_id?`

lib/prediction_analyzer_web/controllers/accuracy_controller.ex
lemald: oops

lib/prediction_analyzer/stop_name_fetcher.ex
lemald: Correct on the name, I'll update that. I'm thinking through what to do about the implementation itself.

lemald: Okay, take a look at the commit I just pushed, I think it should be faster.

PR #92, opened 2019-04-26
checkbox to hide dev green data

sghiaa:
#### Summary of changes
**Asana Ticket:** [[2] analyzer - add checkbox for "include dev green"](https://app.asana.com/0/584764604969369/1116467262379013)

im not sure im happy with the way i did this, but i think doing it another way might be more annoying, and possibly require page reloads.  Originally i wanted to do a thing where we would just take the dev green data off the page entirely, but that would definitely require a page reload and that is nto what we wanted.

#### Reviewer Checklist

### DIFF
--- a/assets/js/chart.js
+++ b/assets/js/chart.js
@@ -7,6 +7,18 @@ export default function() {
       setupDashboard();
     }
 
+    jQuery('#show-dev-green-check').change(function() {
+      var showDevGreen = jQuery("#show-dev-green-check");
+      if (showDevGreen.is(":checked")) {
+        jQuery('#dev-green-data-table').show();
+        jQuery('#dev-green-accuracy-total').show();
+      } else {
+        jQuery('#dev-green-data-table').hide();
+        jQuery('#dev-green-accuracy-total').hide();
+      }
+      setupDashboard();
+    });
+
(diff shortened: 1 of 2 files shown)


assets/js/chart.js
```diff
@@ -7,6 +7,18 @@ export default function() {
       setupDashboard();
     }
 
+    jQuery('#show-dev-green-check').change(function() {
```
phildarnowsky: @sghiaa you can use `$` instead of `jQuery`

phildarnowsky: @sghiaa also it doesn't matter much in this case, but it's best to pass an `event` parameter to the callback and have it prevent the default behavior:

```
$('#show-dev-green-check').change(function(event) {
  event.preventDefault();
  
  // and more code
}

sghiaa: i had been trying iwth $ in the console and it wasnt working, but i guess i didnt check in this file - if i remember correctly they are different contexts, so i could believe it works here.

sghiaa: well i just tried it and no luck.  is there something i should do i.e. $ = jQuery; or something? or is it something that doesnt really matter and i can leave as jQuery?

PR #93, opened 2019-05-01
Ash accuracy drop warn

sghiaa:
#### Summary of changes
**Asana Ticket:** [[3] Analyzer: Alert us if accuracy drops](https://app.asana.com/0/584764604969369/1119496382234059)

when accuracy drops at all between the previous two days, this will log a warning and splunk can be configured to email us.

#### Reviewer Checklist


losvedir:
The ticket mentions the accuracy should be checked by line. But I think this does a single global check?

sghiaa:
it should check every minute right now, cause i wanted to test it first and make sure something happens but i can make it a daily thing right now

test/prediction_analyzer/prediction_accuracy/accuracy_tracker_test.exs
```diff
@@ -0,0 +1,146 @@
...
+        PredictionAccuracy.new_insert_changeset(%{
+          service_date: yesterday,
+          hour_of_day: 5,
+          stop_id: "stop1",
+          route_id: "route1",
+          direction_id: 1,
+          arrival_departure: "arrival",
+          bin: "0-3 min",
+          num_predictions: 100,
+          num_accurate_predictions: 70
```
phildarnowsky: @sghiaa there are enough of these `changeset = ...` lines with virtually the same values that it would pay off to create a helper function where you just pass in the appropriate day and desired `num_accurate_predictions`, for example, in this case you could do something like `changeset = new_changeset(yesterday, 70)`

losvedir: A related approach which we use in a lot of tests is to have one "template" struct as a module attribute:

```ex
@changeset %{
  ...
}
```

and then each test uses the normal elixir map update syntax, to set the keys it cares about:

`%{@changest | service_date: yesterday}`

The advantage is that it's flexible if different tests have different fields they care about (one function would then need lots of parameters). It's not so good, though, if you want to, say, return two related maps.

test/prediction_analyzer/prediction_accuracy/accuracy_tracker_test.exs
```diff
@@ -0,0 +1,146 @@
+defmodule PredictionAnalyzer.PredictionAccuracy.AccuracyTrackerTest do
+  use ExUnit.Case
+  alias PredictionAnalyzer.PredictionAccuracy.AccuracyTracker
+  alias PredictionAnalyzer.PredictionAccuracy.PredictionAccuracy
+  alias PredictionAnalyzer.Repo
+  import ExUnit.CaptureLog
+
+  setup do
+    :ok = Ecto.Adapters.SQL.Sandbox.checkout(PredictionAnalyzer.Repo)
```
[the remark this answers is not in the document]
sghiaa: lets us use the database for the tests

test/prediction_analyzer/prediction_accuracy/accuracy_tracker_test.exs
sghiaa: ok

lib/prediction_analyzer/prediction_accuracy/accuracy_tracker.ex
```diff
@@ -0,0 +1,87 @@
...
+
+    yesterday_accs =
+      yesterday_query
+      |> PredictionAccuracy.stats_by_environment_and_chart_range(%{
+        "chart_range" => "Hourly"
+      })
+      |> PredictionAnalyzer.Repo.all()
+
+    {yesterday_total, yesterday_accurate} =
+      Enum.reduce(yesterday_accs, {0, 0}, fn [_h, p_t, p_a, _, _], {t, a} ->
```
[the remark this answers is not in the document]
sghiaa: fair enough

PR #97, opened 2019-05-21
Route groupings

lemald:
#### Summary of changes
**Asana Ticket:** [[2] Add route groupings to analyzer](https://app.asana.com/0/584764604969369/1120741346356341/f)

Updated `filter_by_route` to call a utility function that converts the text from the parameters to a list of route IDs. For most cases this list will contain only the one route.

#### Reviewer Checklist

### DIFF
--- a/lib/prediction_analyzer/utilities.ex
+++ b/lib/prediction_analyzer/utilities.ex
@@ -91,4 +91,13 @@ defmodule PredictionAnalyzer.Utilities do
   @spec string_to_mode(String.t()) :: :subway | :commuter_rail
   def string_to_mode("commuter_rail"), do: :commuter_rail
   def string_to_mode(_), do: :subway
+
+  @spec route_param_to_list(String.t()) :: [String.t()]
+  def route_param_to_list("Green-All"), do: ["Green-B", "Green-C", "Green-D", "Green-E"]
+
+  def route_param_to_list("Light Rail"),
+    do: ["Green-B", "Green-C", "Green-D", "Green-E", "Mattapan"]
+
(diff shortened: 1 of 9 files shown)


losvedir:
So it seems the idea here is there's a "magic word" that our backend knows how to translate into certain sets of routes.

An alternative approach might be to teach the app to allow `route_id=` to be enumerated with commas (a la the API), and have the view template simply list the route set in the options: `[{"Green-All", "Green-B,Green-C,Green-D,Green-E"}, ...]`.

The advantage would be that we could then query sets of routes without special backend support. For routes I'm not sure it matters much, but for stop_ids I could see it being more relevant.

That was sort of how I was envisioning it, but I see the asana ticket isn't very explicit about it. Given that it's already implemented this way, if it's a lot of work it might not be worth changing it now. But I do think stop_ids should work like that, at least.

lemald:
Hmm, it might not actually be too much work to re-do it that way, and now that you mention it I definitely agree that that's how stop IDs should work. I'll consider it.

lib/prediction_analyzer/utilities.ex
```diff
@@ -91,4 +91,13 @@ defmodule PredictionAnalyzer.Utilities do
   @spec string_to_mode(String.t()) :: :subway | :commuter_rail
   def string_to_mode("commuter_rail"), do: :commuter_rail
   def string_to_mode(_), do: :subway
+
+  @spec route_param_to_list(String.t()) :: [String.t()]
```
losvedir: I don't think this function is used anymore?

lemald: Ah, that is correct, it is not used.

PR #98, opened 2019-06-12
Weekly view

sghiaa:
#### Summary of changes
**Asana Ticket:** [[5] Analyzer: Add a "weekly" view](https://app.asana.com/0/584764604969369/1113375993891945)

im putting this up now because i find it easier to see what ive done nwo that im coming back after a long weekend.  feel free to look if youre interested.

#### Reviewer Checklist

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -30,7 +30,7 @@ config :prediction_analyzer, http_fetcher: HTTPoison
 config :prediction_analyzer, :api_base_url, "https://api-v3.mbta.com/"
 config :prediction_analyzer, :migration_task, Predictions.ReleaseTasks.NoOp
 config :prediction_analyzer, :stop_name_fetcher, PredictionAnalyzer.StopNameFetcher
-config :prediction_analyzer, :timezone, System.get_env("TIMEZONE") || "America/New_York"
+config :prediction_analyzer, :timezone, "America/New_York"
 
 config :prediction_analyzer, PredictionAnalyzer.Repo,
   adapter: Ecto.Adapters.Postgres,
--- a/config/config.exs
+++ b/config/config.exs
@@ -30,6 +30,7 @@ config :prediction_analyzer, http_fetcher: HTTPoison
 config :prediction_analyzer, :api_base_url, "https://api-v3.mbta.com/"
 config :prediction_analyzer, :migration_task, Predictions.ReleaseTasks.NoOp
 config :prediction_analyzer, :stop_name_fetcher, PredictionAnalyzer.StopNameFetcher
+config :prediction_analyzer, :timezone, System.get_env("TIMEZONE") || "America/New_York"
 
 config :prediction_analyzer, PredictionAnalyzer.Repo,
   adapter: Ecto.Adapters.Postgres,
--- a/lib/prediction_analyzer/utilities.ex
+++ b/lib/prediction_analyzer/utilities.ex
@@ -27,6 +27,17 @@ defmodule PredictionAnalyzer.Utilities do
     {date, hour, beginning_of_hour_unix, end_of_hour_unix}
   end
 
+  @spec get_week_range(DateTime.t()) :: {Date.t(), integer(), integer()}
+  def get_week_range(timestamp) do
+    beginning_of_week =
+      timestamp
+      |> Timex.set(hour: 3, minute: 0, second: 0, microsecond: {0, 6})
+
+    end_of_week = Timex.shift(beginning_of_week, days: 7)
+
+    {DateTime.to_date(beginning_of_week), DateTime.to_date(end_of_week)}
+  end
+
   @doc """
   Returns the number of ms to the top of the next hour.
   """
@@ -35,6 +46,21 @@ defmodule PredictionAnalyzer.Utilities do
     local_now
     |> Timex.shift(hours: 1)
     |> Timex.set(minute: 0, second: 30)
+    |> Timex.diff(local_now, :milliseconds)
+  end
+
+  @doc """
+  Returns the ms until end of week
+  """
+  @spec ms_to_next_week(DateTime.t()) :: integer()
+  def ms_to_next_week(local_now \\ Timex.local()) do
+    days_to_end_of_week =
+      local_now
+      |> Timex.days_to_end_of_week()
+
+    local_now
+    |> Timex.shift(days: days_to_end_of_week)
+    |> Timex.set(hour: 1, minute: 0, second: 0)
     |> Timex.diff(local_now, :milliseconds)
   end
 
--- a/lib/prediction_analyzer/utilities.ex
+++ b/lib/prediction_analyzer/utilities.ex
@@ -27,13 +27,13 @@ defmodule PredictionAnalyzer.Utilities do
     {date, hour, beginning_of_hour_unix, end_of_hour_unix}
   end
 
-  @spec get_week_range(DateTime.t()) :: {Date.t(), integer(), integer()}
+  @spec get_week_range(DateTime.t()) :: {Date.t(), Date.t()}
   def get_week_range(timestamp) do
     beginning_of_week =
       timestamp
       |> Timex.set(hour: 3, minute: 0, second: 0, microsecond: {0, 6})
 
-    end_of_week = Timex.shift(beginning_of_week, days: 7)
+    end_of_week = Timex.shift(beginning_of_week, days: 6)
 
     {DateTime.to_date(beginning_of_week), DateTime.to_date(end_of_week)}
   end
@@ -58,10 +58,19 @@ defmodule PredictionAnalyzer.Utilities do
       local_now
       |> Timex.days_to_end_of_week()
 
-    local_now
-    |> Timex.shift(days: days_to_end_of_week)
-    |> Timex.set(hour: 1, minute: 0, second: 0)
-    |> Timex.diff(local_now, :milliseconds)
+    case days_to_end_of_week do
+      0 ->
+        local_now
+        |> Timex.shift(days: 7)
+        |> Timex.set(hour: 1, minute: 0, second: 0)
+        |> Timex.diff(local_now, :milliseconds)
+
+      n ->
+        local_now
+        |> Timex.shift(days: n)
+        |> Timex.set(hour: 1, minute: 0, second: 0)
+        |> Timex.diff(local_now, :milliseconds)
+    end
   end
 
   @doc """
--- a/lib/prediction_analyzer/utilities.ex
+++ b/lib/prediction_analyzer/utilities.ex
@@ -56,7 +56,7 @@ defmodule PredictionAnalyzer.Utilities do
   def ms_to_next_week(local_now \\ Timex.local()) do
     days_to_end_of_week =
       local_now
-      |> Timex.days_to_end_of_week()
+      |> Timex.days_to_end_of_week(:sun)
 
     case days_to_end_of_week do
       0 ->
--- a/lib/prediction_analyzer/weekly_accuracies/aggregator.ex
+++ b/lib/prediction_analyzer/weekly_accuracies/aggregator.ex
@@ -19,7 +19,7 @@ defmodule PredictionAnalyzer.WeeklyAccuracies.Aggregator do
       |> Timex.days_to_end_of_week()
 
     backfill_start =
-      case Map.get(opts, :backfill_start, nil) do
+      case Map.get(opts, :backfill_time, nil) do
         nil ->
           local_now
           |> Timex.shift(days: days_to_end_of_week)
--- a/lib/prediction_analyzer/weekly_accuracies/aggregator.ex
+++ b/lib/prediction_analyzer/weekly_accuracies/aggregator.ex
@@ -3,11 +3,11 @@ defmodule PredictionAnalyzer.WeeklyAccuracies.Aggregator do
   require Logger
   alias PredictionAnalyzer.WeeklyAccuracies.Query
 
-  def start_link(opts \\ []) do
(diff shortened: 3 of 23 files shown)


lib/prediction_analyzer/utilities.ex
```diff
@@ -27,6 +27,17 @@ defmodule PredictionAnalyzer.Utilities do
     {date, hour, beginning_of_hour_unix, end_of_hour_unix}
   end
 
+  @spec get_week_range(DateTime.t()) :: {Date.t(), integer(), integer()}
```
lemald: I don't know if this is what Semaphore is complaining about, but I think this should be `{DateTime.t(), DateTime.t()}`.

lib/prediction_analyzer/utilities.ex
```diff
@@ -27,6 +27,17 @@ defmodule PredictionAnalyzer.Utilities do
...
+  def get_week_range(timestamp) do
+    beginning_of_week =
+      timestamp
+      |> Timex.set(hour: 3, minute: 0, second: 0, microsecond: {0, 6})
```
lemald: I'm confused from reading this, how does this get the first day of the week? It looks to me like it would just get 3:00 AM on the day of `timestamp`.

sghiaa: thats what it does, and the end of the week is 6 days later.

lib/prediction_analyzer/weekly_accuracies/aggregator.ex
```diff
@@ -0,0 +1,70 @@
...
+
+  def start_link(opts \\ []) do
+    GenServer.start_link(__MODULE__, [], opts)
+  end
+
+  def init([]) do
+    schedule_next_week(self())
+    schedule_backfill(self())
+
+    local_now = Timex.now("America/New_York")
```
lemald: It looks like there are a couple of places in the repo that have `America/New_York` hardcoded, would it be hard to put the time zone in a config variable?

sghiaa: i think this decision was made because our predictions and analyzer are always in america/new_york and we didnt want it to change.

lib/prediction_analyzer/utilities.ex
lemald: Okay, so `ms_to_next_week` determines when we run the aggregation, and then `get_week_range` will always get called at the beginning on the week?

lib/prediction_analyzer/weekly_accuracies/aggregator.ex
lemald: Hmm, I don't personally remember making this decision and I just chatted with Gabe and he doesn't remember it either. I know we don't expect it to change but we've also discussed Prediction Analyzer as a decent candidate to open-source in the future, in which case we would definitely want it to be configurable.

Also for what it's worth RTR has its time zone in the configuration, and that's probably even less likely to be portable.

sghiaa: thats interesting that he doesnt remember, the only reason we did it was because he had issues where we were getting his local timezone.  either way, i can change it

config/config.exs
```diff
@@ -30,6 +30,7 @@ config :prediction_analyzer, http_fetcher: HTTPoison
 config :prediction_analyzer, :api_base_url, "https://api-v3.mbta.com/"
 config :prediction_analyzer, :migration_task, Predictions.ReleaseTasks.NoOp
 config :prediction_analyzer, :stop_name_fetcher, PredictionAnalyzer.StopNameFetcher
+config :prediction_analyzer, :timezone, System.get_env("TIMEZONE") || "America/New_York"
```
lemald: Oh, I think what I had in mind was just having it set here as `America/New_York`, not using the `TIMEZONE` system environment variable. Now I understand why Gabe might have had an issue with this.

lib/prediction_analyzer/weekly_accuracies/aggregator.ex
```diff
@@ -0,0 +1,72 @@
...
+    Query.calculate_weekly_accuracies(
+      PredictionAnalyzer.Repo,
+      state[:backfill_time]
+    )
+
+    next_backfill =
+      state[:backfill_time]
+      |> Timex.shift(days: -7)
+
+    schedule_backfill(self())
```
lemald: Does this ever stop going back indefinitely? Obviously it will eventually stop getting data, but it looks to me like it will continue hitting `schedule_backfill` on another 1-week period further in the past.

lemald: Oh, also, it might be nice to include some logging.

sghiaa: it doesnt stop because i couldnt think of a reliable way that wed want it to stop.  i think id want to do a followup to just remove the backfill at some point.

what would trigger us to stop backfilling?  my best thought was a week with no data, but i dont know if we can guarantee that we have data for every week, and even if we can it doesnt really hurt to keep going back.  

logging is a good idea.

lib/prediction_analyzer/weekly_accuracies/weekly_accuracies.ex
```diff
@@ -0,0 +1,209 @@
+defmodule PredictionAnalyzer.WeeklyAccuracies.WeeklyAccuracies do
```
lemald: A lot of the helper functions in this module are quite similar to those in the `PredictionAccuracy` module, is there any way to put those into a shared place?

sghiaa: yeah i thought about that at the time and then forgot about it because this ticket has taken a million years.  ill think about it again and hopefully this time do something about it.

lib/prediction_analyzer/weekly_accuracies/aggregator.ex
lemald: Hmm, fair point. If it's temporary you could maybe hardcode in some stop date from like a month ago?

One other thing that occurred to me, since you set `backfill_time` on the init of this Genserver, doesn't that mean that every time the app restarts it will start backfilling from whatever the current date is? Were you planning on just reverting the backfill logic right after deploying and letting it do its thing?

sghiaa: it will start checking with the current week of whenever we restart, but it wont actually do anything if its already aggregated that data.  

on the other point, we dont want to stop a month ago, this is aggregating the aggregated data so it can go back to the first aggregated day.

lemald: Ah, right, good point about how this is going over the aggregated data and not the original data. And on the point of re-aggregating weeks we've already done, I forgot that it checks before doing the actual aggregation, so I think this is fine.

lib/prediction_analyzer/weekly_accuracies/weekly_accuracies.ex
```diff
@@ -0,0 +1,209 @@
...
+      {:ok, from(acc in q, where: acc.bin == ^bin)}
+    else
+      {:ok, q}
+    end
+  end
+
+  @spec filter_by_timeframe(Ecto.Query.t(), any(), any(), any(), any()) ::
+          {:ok, Ecto.Query.t()} | {:error, String.t()}
+  defp filter_by_timeframe(q, chart_range, _date, start_date, end_date)
+       when chart_range == "Weekly" and is_binary(start_date) and is_binary(end_date) do
```
lemald: This can probably be simplified if `chart_range` is only ever going to be "weekly" in this case (though I guess that might also depend on if / how you're able to combine these functions with their counterparts in `prediction_accuracy.ex`).

lib/prediction_analyzer/weekly_accuracies/weekly_accuracies.ex
```diff
@@ -0,0 +1,209 @@
...
+  Takes a Queryable and groups and sums the results into
+  a table like:
+
+  hour | prod total | prod accurate | dev-green total | dev-green accurate
+  """
+  @spec stats_by_environment_and_chart_range(Ecto.Query.t(), map()) :: Ecto.Query.t()
+  def stats_by_environment_and_chart_range(q, filters) do
+    scope =
+      case filters["chart_range"] do
+        "Weekly" -> :week_start
```
lemald: This part can probably also be simplified if the range is only ever going to be weekly.

lib/prediction_analyzer_web/controllers/accuracy_controller.ex
```diff
@@ -1,10 +1,72 @@
 defmodule PredictionAnalyzerWeb.AccuracyController do
   use PredictionAnalyzerWeb, :controller
   alias PredictionAnalyzer.PredictionAccuracy.PredictionAccuracy
+  alias PredictionAnalyzer.WeeklyAccuracies.WeeklyAccuracies
 
   import Ecto.Query, only: [from: 2]
 
   @spec index(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def index(
```
lemald: Similar to how there's some similarity between `WeeklyAccuracy` and `PredictionAccuracy`, I wonder if this clause of `index` could be reasonably combined with the following one, calling the appropriate versions of `filter` and `stats_by_environment_and_chart_range` based on `chart_range`.

lib/prediction_analyzer/weekly_accuracies/weekly_accuracies.ex
sghiaa: i didnt move the `filter` functions because it seemed like it made sense as part of the module with the schema, but now that im looking at it again im not sure i agree with myself.  what do you think

lemald: My gut instinct is that leaving them in the respective schema modules looks fine.

lib/prediction_analyzer/weekly_accuracies/aggregator.ex
```diff
@@ -0,0 +1,80 @@
...
+          current_time
+        )
+      end)
+
+    Logger.info("Finished weekly prediction aggregations in #{time / 1000} ms")
+    schedule_next_week(self())
+    {:noreply, state}
+  end
+
+  defp schedule_backfill(pid) do
```
lemald: This and `schedule_next_week` should get typespecs.

lib/prediction_analyzer/weekly_accuracies/weekly_accuracies.ex
```diff
@@ -0,0 +1,62 @@
...
+    field(:stop_id, :string)
+    field(:route_id, :string)
+    field(:direction_id, :integer)
+    field(:arrival_departure, :string)
+    field(:bin, :string)
+    field(:num_predictions, :integer)
+    field(:num_accurate_predictions, :integer)
+  end
+
+  def new_insert_changeset(params \\ %{}) do
```
lemald: Typespec

lib/prediction_analyzer/utilities.ex
```diff
@@ -38,6 +49,30 @@ defmodule PredictionAnalyzer.Utilities do
...
   end
 
+  @doc """
+  Returns the ms until end of week
+  """
+  @spec ms_to_next_week(DateTime.t()) :: integer()
+  def ms_to_next_week(local_now \\ Timex.local()) do
+    days_to_end_of_week =
+      local_now
+      |> Timex.days_to_end_of_week()
```
lemald: Okay, one minor thing: the ticket specifies that weeks should start on Sunday, but it looks like Monday is the default beginning-of-week day that Timex uses. It looks like you should just be able to specify `:sun` here in the second argument, though, and any other places where `Timex.days_to_end_of_week` is used.

lib/prediction_analyzer_web/controllers/accuracy_controller.ex
```diff
@@ -1,10 +1,73 @@
...
+        chart_data: Jason.encode!(set_up_accuracy_chart(accuracies, filter_params)),
+        prod_num_accurate: prod_num_accurate,
+        prod_num_predictions: prod_num_predictions,
+        dev_green_num_accurate: dev_green_num_accurate,
+        dev_green_num_predictions: dev_green_num_predictions,
+        error_msg: error_msg,
+        mode: mode_atom
+      )
+    else
+      redirect_with_default_filters(conn, params)
```
lemald: I think `redirect_with_default_filters` still needs to handle the default filters for the weekly view (last 10 weeks).

lib/prediction_analyzer/utilities.ex
sghiaa: oh true

lib/prediction_analyzer_web/controllers/accuracy_controller.ex
sghiaa: hmm that must have been lost at some point

test/prediction_analyzer_web/controllers/accuracy_controller_test.exs
```diff
@@ -26,7 +26,7 @@ defmodule PredictionAnalyzerWeb.AccuracyControllerTest do
     environment: "prod",
     num_accurate_predictions: 327,
     num_predictions: 617,
-    route_id: "Blue",
+    route_id: "blue",
```
lemald: Any particular reason for this change, or just a typo?

sghiaa: theres a vim keybind that lowercases a selection that is very similar to copy a selection. so, typo

sghiaa:
coolio, willd o

PR #99, opened 2019-06-26
Add ME and RMSE columns and calculations

losvedir:
#### Summary of changes
**Asana Ticket:** Part of [📈  Add average deviation as a metric](https://app.asana.com/0/584764604969369/1113586286577022)

Adds the mean error and root mean squared error to our predictions. A subsequent PR will expose this data in the front end.

One little tricky bit I worked the math out for that I'd like double checked is how to aggregate multiple errors / rmses.

Mean error:

If 5 predictions had a mean error of 10 and 2 predictions had a mean error of 8, then the overall mean error is `(5*10 + 2*8) / (5+2)`.

RMSE:

If 5 predictions had an RMSE of 10 and 2 predictions had an RMSE of 8, then the overall RMSE is: 
`sqrt((5*10^2 + 2*8^2) / (5+2))`.

#### Reviewer Checklist

### DIFF
--- a/priv/repo/migrations/20190624192925_add_mse_rmse_columns.exs
+++ b/priv/repo/migrations/20190624192925_add_mse_rmse_columns.exs
@@ -3,13 +3,13 @@ defmodule PredictionAnalyzer.Repo.Migrations.AddMseRmseColumns do
 
   def change do
     alter table("prediction_accuracy") do
-      add(:mean_error, :real, null: false, default: 0.0)
-      add(:root_mean_squared_error, :real, null: false, default: 0.0)
+      add(:mean_error, :real)
+      add(:root_mean_squared_error, :real)
     end
 
     alter table("weekly_accuracies") do
-      add(:mean_error, :real, null: false, default: 0.0)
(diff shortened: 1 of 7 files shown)


sghiaa:
according to me and a piece of paper working it out independently of what you wrote, I agree with these formulae

priv/repo/migrations/20190624192925_add_mse_rmse_columns.exs
```diff
@@ -0,0 +1,15 @@
+defmodule PredictionAnalyzer.Repo.Migrations.AddMseRmseColumns do
```
sghiaa: we can have a downgrade for thse too.  `def down` i think?

losvedir: `change` will automatically generate the `up` and `down`. I just did a rollback locally and it was able to infer the columns to drop.

sghiaa: well thatts clever

PR #100, opened 2019-06-28
Show mean error and RMSE on frontend.

losvedir:
#### Summary of changes
**Asana Ticket:** Completes [📈  Add average deviation as a metric](https://app.asana.com/0/584764604969369/1113586286577022).

The DB has already been updated and has been collecting the data. This PR exposes it on the frontend. The commits are self-contained and can be reviewed sequentially:

1) A simple refactor where two huge functions were almost identical except for one line. I consolidated them.

2) Renamed a file `filter.ex` -> `filters.ex` since the module was `Filters`.

3) Calculate the overall mean error and RMSE and put them in the top line with the prod and dev-green summaries. Since this calculation will be used later, it's done as a macro that can be imported and used in the Ecto query macro.

Screen Shot 2019-06-28 at 1 38 55 PM

4) Add the Error/RMSE to the table rows below the chart.

Screen Shot 2019-06-28 at 1 39 01 PM

[Please include a brief description of what was changed]

#### Reviewer Checklist


test/prediction_analyzer/query_utilities_test.exs
```diff
@@ -0,0 +1,191 @@
...
+    bin: "0-3 min",
+    num_predictions: 1,
+    num_accurate_predictions: 1,
+    mean_error: 10.0,
+    root_mean_squared_error: 10.0
+  }
+
+  setup do
+    :ok = Ecto.Adapters.SQL.Sandbox.checkout(PredictionAnalyzer.Repo)
+    # Ecto.Adapters.SQL.Sandbox.mode(PredictionAnalyzer.Repo, {:shared, self()})
```
lemald: Can probably get rid of this if it's commented out.

lib/prediction_analyzer_web/templates/accuracy/index.html.eex
```diff
@@ -32,6 +33,7 @@
       </span>
     </div>
     <div class="dev-green-text">
+      <em>Err: <%= if @dev_green_mean_error, do: Float.round(@dev_green_mean_error, 2) %>, RMSE: <%= if @dev_green_rmse, do: Float.round(@dev_green_rmse, 2) %></em> <br>
```
lemald: Why do the `if @dev_green_mean_error, do: ...` when you don't do the similar conditional for the prod errors?

losvedir: Oh, oops. I missed that one. I had them all as `if ..., do:` before I realized that our convention seemed to be to have `0.0` as the default rather than the empty string (e.g. if you load a date before we have data, you'll see the accuracy as 0%). So elsewhere I changed it to `Float(val || 0.0, 2)`, rather than `if val, do: Float(val, 2)`. But I missed it here. I'll update.

PR #101, opened 2019-07-02
Split 6-12 min bucket into more granularity

losvedir:
#### Summary of changes
**Asana Ticket:** [Split 6-12 bin to 6-8, 8-10, 10-12](https://app.asana.com/0/584764604969369/1128108664093986)

Splits the 6-12 min bucket apart into 6-8, 8-10, and 10-12 mins.

In postgres, this column was an enum, so it requires a migration to alter the type to add the new values. This can't be run in a transaction so I had to use `@disable_ddl_transaction`. Unfortunately, there's no good way to _remove_ values from a postgres type enum, so while the `up/0` adds them, the `down/0` does nothing. The work around seems to be to make a new type, and alter all the columns with the old type to the new one, then delete the old type, and then rename the new type. But given we'd be dealing with millions of rows, that's not something I want to have done automatically and nonchalantly in a migration rollback. If there's an issue with the migration for some reason, we can tackle how to solve it specifically.

The aggregation code change was very small. Updating `bins/0` to specify the new bins and remove the old one.

I updated the query to take a comma separated list of bins, a la route_id, and then added a new option to the bin select "6-12 min" which has a value of "6-8 min,8-10 min,10-12 min,6-12 min", so that the "6-12 bin" is comparable before and after the split.

#### Reviewer Checklist

### DIFF
--- a/priv/repo/migrations/20190701174220_add_buckets_to_accuracy_tables.exs
+++ b/priv/repo/migrations/20190701174220_add_buckets_to_accuracy_tables.exs
@@ -1,0 +1,15 @@
+defmodule PredictionAnalyzer.Repo.Migrations.AddBucketsToAccuracyTables do
+  use Ecto.Migration
+  # adding value to enum type can't be done in a transaction
+  @disable_ddl_transaction true
+
+  def up do
+    execute("alter type prediction_bin add value '6-8 min'")
+    execute("alter type prediction_bin add value '8-10 min'")
+    execute("alter type prediction_bin add value '10-12 min'")
+  end
+
+  def down do
+    # can't easily remove an enum type value
+  end
+end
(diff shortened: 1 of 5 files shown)


priv/repo/migrations/20190701174220_add_buckets_to_accuracy_tables.exs
```diff
@@ -0,0 +1,15 @@
...
+  @disable_ddl_transaction true
+
+  def up do
+    execute("alter type prediction_bin add value '6-8 min'")
+    execute("alter type prediction_bin add value '8-10 min'")
+    execute("alter type prediction_bin add value '10-12 min'")
+  end
+
+  def down do
+    # can't easily remove an enum type value
```
phildarnowsky: @losvedir do we maybe want to raise an error here, the way you might throw an `IrreversibleMigration` exception in Rails? That way if you happen to roll this migration back you're not left with the false impression that the rollback actually did anything.

losvedir: Good idea. I'll raise an `Ecto.MigrationError` which is raised in other non-reversible scenarios.

PR #104, opened 2019-07-15
run the pruner more often

sghiaa:
#### Summary of changes
**Asana Ticket:** [💳 Prediction analyzer pruning failing](https://app.asana.com/0/584764604969369/1130638623059531)

pruner ran every day at 3am before.  now every 12 hours. we will also need to run the delete query live at some point.
#### Reviewer Checklist

### DIFF
--- a/lib/prediction_analyzer/pruner.ex
+++ b/lib/prediction_analyzer/pruner.ex
@@ -25,7 +25,6 @@ defmodule PredictionAnalyzer.Pruner do
     unix_cutoff =
       Timex.local()
       |> Timex.shift(days: -28)
-      |> Timex.set(hour: 3, minute: 0, second: 0)
       |> DateTime.to_unix()
 
     {time, _} =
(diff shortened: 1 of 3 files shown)


lib/prediction_analyzer/pruner.ex
```diff
@@ -68,6 +67,6 @@ defmodule PredictionAnalyzer.Pruner do
   end
 
   defp schedule_next_run(pid) do
-    Process.send_after(pid, :prune, Utilities.ms_to_3am(Timex.local()) + 10 * 60 * 1_000)
+    Process.send_after(pid, :prune, (Timex.local() |> DateTime.to_unix()) + 12 * 60 * 60 * 1_000)
```
lemald: Two things:
1. Is there anything else that uses `ms_to_3am`? I think this might be the only thing that invoked it, in which case you can just delete it.
2. Typespec for `schedule_next_run`.

sghiaa: typespecing private functions that only do Process.send_after might be my least favorite thing we do

PR #111, opened 2019-09-19
Log case where we fail to parse a given vehicle

lemald:
#### Summary of changes
**Asana Ticket:** [🐛 Prediction Analyzer "multiple departure" errors](https://app.asana.com/0/584764604969369/1125718067018656/f)

It looks like in many of these cases the vehicles are getting "materialized" as new, but in the concrete examples I've looked at I can't find any reason why they wouldn't be in the vehicle positions feed for an update. The one thing I can think of is if the exact structure of the vehicle entity somehow fails to match the function clause in the `PredictionAnalyzer.VehiclePositions.Vehicle.from_json` function. Logging this will allow me to test this theory.

#### Reviewer Checklist

### DIFF
--- a/lib/prediction_analyzer/vehicle_positions/tracker.ex
+++ b/lib/prediction_analyzer/vehicle_positions/tracker.ex
@@ -156,8 +156,12 @@ defmodule PredictionAnalyzer.VehiclePositions.Tracker do
   defp parse_vehicles(%{"entity" => entities}, environment) do
     Enum.flat_map(entities, fn e ->
       case Vehicle.from_json(e, environment) do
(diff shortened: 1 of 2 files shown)


losvedir:
That's a clever idea. Did you try running it locally at all and see if it caught anything?

lemald:
@losvedir so far all I've seen is instances where it looks like the failure is due to the stop_id being nil, which I don't think is causing the issues we're seeing. I hope to look at some concrete "one departure, multiple updates" errors and see if I can tie them to this logging, which will probably be easiest if it runs in the actual prod environment for a while.

PR #120, opened 2019-12-18
fix: Don't [warn] for expected vehicles without stop IDs

losvedir:
#### Summary of changes
**Asana Ticket:** [[extra] Don't warn in prediction analyzer when parsing vehicles without stop IDs, which are expected](https://app.asana.com/0/584764604969369/1154516134595690)

I was noticing we had a lot of warnings in our logs from this, but it turns out vehicles without stop IDs are expected from time to time, so we shouldn't warn for them.

#### Reviewer Checklist

### DIFF
--- a/lib/prediction_analyzer/vehicle_positions/tracker.ex
+++ b/lib/prediction_analyzer/vehicle_positions/tracker.ex
@@ -153,11 +153,15 @@ defmodule PredictionAnalyzer.VehiclePositions.Tracker do
     {:noreply, state}
   end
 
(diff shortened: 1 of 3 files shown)


lib/prediction_analyzer/vehicle_positions/tracker.ex
```diff
@@ -159,6 +159,9 @@ defmodule PredictionAnalyzer.VehiclePositions.Tracker do
         {:ok, vehicle} ->
           [vehicle]
 
+        :ignore ->
+          []
```
lemald: Typespec for `parse_vehicles`.

PR #306, opened 2020-10-15
chore: upgrade Phoenix to 1.5.0

lemald:
#### Summary of changes
**Asana Ticket:** [💳 upgrade prediction_analyzer to Phoenix 1.5](https://app.asana.com/0/584764604969369/1191831644615745/f)

A lot of Ecto stuff is also out of date, but upgrading that is causing some test failures where it isn't obvious to me exactly what's going wrong.

#### Reviewer Checklist


lib/prediction_analyzer/application.ex
```diff
@@ -6,40 +6,37 @@ defmodule PredictionAnalyzer.Application do
...
     supervisors = [
       # Start the Ecto repository
-      supervisor(PredictionAnalyzer.Repo, []),
+      PredictionAnalyzer.Repo,
       # Start the endpoint when the application starts
-      supervisor(PredictionAnalyzerWeb.Endpoint, [])
-      # Start your own worker by calling: PredictionAnalyzer.Worker.start_link(arg1, arg2, arg3)
-      # worker(PredictionAnalyzer.Worker, [arg1, arg2, arg3]),
+      {Phoenix.PubSub, name: PredictionAnalyzerWeb.PubSub},
+      PredictionAnalyzerWeb.Endpoint
```
digitalcora, line 16: I think this comment
```elixir
# Start the endpoint when the application starts
```
is intended to refer to this line
```elixir
PredictionAnalyzerWeb.Endpoint
```
rather than the line it now finds itself above. Granted, I don't know if we really need these comments at all, since the setup is pretty conventional for a Phoenix app.

lib/prediction_analyzer_web/templates/layout/app.html.eex
```diff
@@ -22,7 +22,7 @@
 
           Prediction analyzer
         </h1>
-        <%= render @view_module, @view_template, assigns %>
+        <%= @inner_content %>
```
digitalcora, line 25: Nice!

digitalcora (approved):
Looks good! Had one tiny comment, not blocking.

lib/prediction_analyzer/application.ex
lemald: Ah, good point. I'll just get rid of the comments here, they're not exactly terribly informative.

PR #307, opened 2020-10-16
Add direnv and dev database improvements

digitalcora:
**Asana Ticket:** [📈 [extra] use `direnv`](https://app.asana.com/0/584764604969369/1198494854059527)

Fell down a database configuration hole after adding `direnv` and realizing it didn't quite work with how the app expected `DATABASE_URL` to be used, but figured out a reasonable way forward. I considered changing how the database is set up in production instead (e.g. to be more like Arrow), but I think we will end up revisiting this when we upgrade to Ecto 3 and/or when we terraformize the app.

See commit messages for details!

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -12,10 +12,7 @@ config :logger, level: :warn
 
 # Configure your database
 config :prediction_analyzer, PredictionAnalyzer.Repo,
-  username: System.get_env("DATABASE_POSTGRESQL_USERNAME") || "postgres",
-  password: System.get_env("DATABASE_POSTGRESQL_PASSWORD") || "postgres",
-  database: "prediction_analyzer_test",
-  hostname: "localhost",
+  url: "#{System.get_env("DATABASE_ROOT_URL")}/prediction_analyzer_test",
   pool: Ecto.Adapters.SQL.Sandbox
 
 config :prediction_analyzer,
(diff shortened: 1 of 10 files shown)


config/test.exs
```diff
@@ -12,10 +12,7 @@ config :logger, level: :warn
 
 # Configure your database
 config :prediction_analyzer, PredictionAnalyzer.Repo,
-  username: System.get_env("DATABASE_POSTGRESQL_USERNAME") || "postgres",
-  password: System.get_env("DATABASE_POSTGRESQL_PASSWORD") || "postgres",
-  database: "prediction_analyzer_test",
-  hostname: "localhost",
+  url: "#{System.get_env("DATABASE_ROOT_URL")}/prediction_analyzer_test",
```
losvedir, line 15: I was sort of surprised this worked, but it looks like you added this to Semaphore. However, it seems to hardcode the username/password. Where did you get that from, and is it guaranteed to remain static?

When we set this up initially, per Semaphore's recommendation we used `DATABASE_POSTGRESQL_USERNAME` and `_PASSWORD`, which they populate.

If the username/password can change, maybe it would be better to `export DATABASE_URL` in `semaphore/setup.sh` based on the `_USERNAME` and `_PASSWORD` that are provided by Semaphore?

digitalcora: > However, it seems to hardcode the username/password. Where did you get that from, and is it guaranteed to remain static?

[The Postgres docs](https://classic-docs.semaphoreci.com/docs/databases/postgresql.html) list the values as well as the keys, which certainly makes them seem static. Notably, in the docs for the [newer platform version](https://semaphoreci.com/docs/ubuntu-1804.html#databases), the environment variables aren't even mentioned — it just says:

> The default PostgreSQL and PostGIS username is `runner` and the password is `semaphoredb`.

So, I don't think these credentials are expected to change, at least within a given platform version.

losvedir (approved):
👍

PR #314, opened 2020-10-21
feat: add multi-select and groups to stop filter

digitalcora:
The stop filter is now a `<select multiple>` and has the conventional behavior for submitting multiple values for the same form field, i.e. the URL is `filters[stop_ids][]=70001&filters[stop_ids][]=...` rather than `filters[stop_id]=70001`.

Our existing `bootstrap-combobox` library didn't support multi-select, so this swaps it out for Selectize. Unfortunately this library is not actively maintained (last release in 2006), but it is otherwise stable and one of the few that doesn't require React/Vue/etc., works correctly out of the box with our Webpack/jQuery/Bootstrap setup, and has search ranking (i.e. typing "Wonderland" puts the station itself at the top, above all the stops that have "Wonderland" elsewhere in their name and would be sorted above it alphabetically).

To avoid creating very long URLs, stop groups are implemented as single "special values" (such as `_terminal`) on the client side, which the server translates into a set of stop IDs. For now these are hard-coded.

**Asana Ticket:** [📈 add stop groupings to analyzer](https://app.asana.com/0/584764604969369/1120764604279515)

Screen-Recording-2020-10-21-at-3

#### Reviewer Checklist

### DIFF
--- a/lib/prediction_analyzer_web/views/accuracy_view.ex
+++ b/lib/prediction_analyzer_web/views/accuracy_view.ex
@@ -107,26 +107,24 @@ defmodule PredictionAnalyzerWeb.AccuracyView do
     false
   end
 
-  @spec stop_descriptions(PredictionAnalyzer.Utilities.mode()) :: [{String.t(), String.t()}]
-  def stop_descriptions(mode) do
-    stop_name_fetcher = Application.get_env(:prediction_analyzer, :stop_name_fetcher)
-
-    description_pairs =
-      stop_name_fetcher.get_stop_descriptions(mode)
-      |> Enum.map(&stop_description/1)
+  @spec stop_filter_options(PredictionAnalyzer.Utilities.mode()) ::
+          %{String.t() => [{String.t(), String.t()}]} | [{String.t(), String.t()}]
+  def stop_filter_options(mode) do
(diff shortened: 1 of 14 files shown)


lib/prediction_analyzer_web/views/accuracy_view.ex
```diff
@@ -107,27 +107,25 @@ defmodule PredictionAnalyzerWeb.AccuracyView do
...
-  defp stop_description({id, description}) do
-    if description do
-      {"#{description} (#{id})", id}
-    else
-      {id, id}
+    # Groups are currently not applicable to Commuter Rail
+    case mode do
+      :commuter_rail -> stop_options
+      :subway -> %{"Groups" => Filters.StopGroups.group_names(), "Stops" => stop_options}
     end
```
lemald, line 122: Rather than having two separate cases, could you combine them and have `:commuter_rail` just return `%{"Groups" => [], "Stops" => stop_options}` or something?

lemald: Edit: never mind, I now understand the purpose of the map vs. list.

lemald (approved):
LGTM!

PR #315, opened 2020-10-21
Improve matching of vehicle events, other small fixes

lemald:
#### Summary of changes
**Asana Ticket:** [📈 investigate: improving handling of times](https://app.asana.com/0/584764604969369/1196078640623588/f)

Previously, when recording the arrival time of a vehicle event, we would match based on the trip ID of the vehicle. However, if this changed mid-route, we could get artificially penalized.

A [followup ticket](https://app.asana.com/0/881264583703207/1198881667597053/f) was created for the more involved changes to how aggregation works, since it might change things in ways that makes it hard to compare future data to past data.

There are a couple of other small commits in here fixing other things as well, see full commit messages for details.

#### Reviewer Checklist

### DIFF
--- a/test/prediction_analyzer/vehicle_positions/comparator_test.exs
+++ b/test/prediction_analyzer/vehicle_positions/comparator_test.exs
@@ -44,7 +44,6 @@ defmodule PredictionAnalyzer.VehiclePositions.ComparatorTest do
 
   describe "compare_vehicles" do
     test "records arrival and departure of vehicle" do
-      Logger.configure(level: :info)
       assert Repo.one(from(ve in VehicleEvent, select: count(ve.id))) == 0
 
       old_vehicles = %{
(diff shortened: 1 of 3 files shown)


test/prediction_analyzer/vehicle_positions/comparator_test.exs
```diff
@@ -44,7 +44,6 @@ defmodule PredictionAnalyzer.VehiclePositions.ComparatorTest do
 
   describe "compare_vehicles" do
     test "records arrival and departure of vehicle" do
-      Logger.configure(level: :info)
```
digitalcora, line 47: I'm curious whether this was the only thing stopping this test module from being `async`.

digitalcora (approved):
Good stuff! 🉑 Matching with the vehicle ID def makes more sense.

test/prediction_analyzer/vehicle_positions/comparator_test.exs
lemald: Unfortunately there's still a bunch of other stuff that uses `capture_log` in here, and I think that `capture_log` will also capture things from other tests running asynchronously.

PR #316, opened 2020-10-22
perf: adjust database indexes for usage patterns

digitalcora:
**Asana Ticket:** [📈 improve indexes and matching queries](https://app.asana.com/0/584764604969369/1195680449065004)

This makes the following index changes:

* Drop `trip_id` on `predictions`, since this index is seemingly unused (both in the code and as indicated by Postgres stats), and is taking up ~11GB of space that could be reclaimed.

* Add `stop_id` on `predictions`, since this column is queried in the process of associating vehicle events with predictions. The index is partial on rows where `vehicle_event_id IS NULL`, since this condition is always included in the queries that include `stop_id`. This should save a bit of space since the condition is only true for half the rows in the current production database (the now-dropped `trip_id` index did the same thing).

* Replace `vehicle_id` on `predictions` with a partial index having the same condition as the above, for the same reason.

* Add `stop_id` and `vehicle_id` on `vehicle_events`, partial on rows where `departure_time IS NULL`, since these columns are queried when filling in departure times.

* Add `stop_id` on `prediction_accuracy`, since filtering on this field is part of the frontend's regular operation.

Also considered but *not* implemented were indexes on `predictions` for `route_id`, and on `prediction_accuracy` for `route_id` and `bin`. While we do query on these fields often, it's always in combination with at least one other indexed field; and since there are only a few distinct routes and bins, the index likely wouldn't narrow the result set enough to justify its use (to the query planner, if not to us).

### Notes

We drop and create all indexes concurrently, which enables the app to continue querying and updating the tables being indexed, at the cost of making the indexing itself slower. Each `CREATE/DROP INDEX` statement still blocks until the index is created/dropped, but the migrate-on-startup behavior runs the migrations inside a `spawn/1`, so this should not prevent the app from starting up. Unfortunately it does mean if the app instance running the migration goes down for whatever reason, the migration state (and maybe the index state) will be left inconsistent, since concurrent index operations can't be done in a transaction. The impact is just that it will take some extra work to clean it up manually — the app should continue to function in any case.

### Reviewer Checklist

### DIFF
--- a/priv/repo/migrations/20201022151038_adjust_indexes.exs
+++ b/priv/repo/migrations/20201022151038_adjust_indexes.exs
@@ -1,0 +1,22 @@
+defmodule PredictionAnalyzer.Repo.Migrations.AdjustIndexes do
+  use Ecto.Migration
+  @disable_ddl_transaction true
+
+  def change do
+    drop(index(:predictions, :trip_id, where: "vehicle_event_id IS NULL", concurrently: true))
+    drop(index(:predictions, :vehicle_id, concurrently: true))
+    create(index(:vehicle_events, :stop_id, where: "departure_time IS NULL", concurrently: true))
+
+    create(
+      index(:vehicle_events, :vehicle_id, where: "departure_time IS NULL", concurrently: true)
+    )
+
+    create(index(:predictions, :stop_id, where: "vehicle_event_id IS NULL", concurrently: true))
+
+    create(
+      index(:predictions, :vehicle_id, where: "vehicle_event_id IS NULL", concurrently: true)
+    )
+
+    create(index(:prediction_accuracy, :stop_id, concurrently: true))
+  end
+end
--- a/priv/repo/structure.sql
+++ b/priv/repo/structure.sql
@@ -246,6 +246,13 @@ CREATE INDEX prediction_accuracy_service_date_index ON public.prediction_accurac
 
 
 --
+-- Name: prediction_accuracy_stop_id_index; Type: INDEX; Schema: public; Owner: -
+--
+
+CREATE INDEX prediction_accuracy_stop_id_index ON public.prediction_accuracy USING btree (stop_id);
+
+
+--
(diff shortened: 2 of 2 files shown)


losvedir:
> Drop `trip_id` on `vehicle_events`, since as of `19c9ef5` this column is no longer the subject of any query in the app, and the index is taking up ~10GB of space that could be reclaimed.

This doesn't seem plausible to me since `vehicle_events` only has ~75k rows. However, the code looks like it drops it from `predictions` which _would_ make sense.

I just wanted to call this out rather than assume typo, though, since the commit that you linked to does actually refer to a query using `trip_id` on `vehicle_events`. I just wanted to double check that your analysis on this removal still applies.

digitalcora:
Thanks, I made a copy-paste error here and meant to refer to the `trip_id` index on `predictions`. Not sure why I associated Eddie's commit with it; I knew that it removed a reference to `trip_id` so I probably misremembered the table name and didn't look closely enough.

losvedir:
These changes look good to me and the reasoning sounds plausible. Once we deploy them we can compare the query times and see if the postgres query planner decides to use them.

I'm going to hold off on the approval for now, though, since this will auto-deploy with merge, and it's the kind of thing I'd rather see happen on Monday.

But here's a non-approval 👍 for now!

digitalcora:
> Once we deploy them we can compare the query times and see if the postgres query planner decides to use them.

Yes, I was thinking we might not necessarily keep all these in the future, since the extra work of combining multiple indexes sometimes isn't worth it — and they are _very_ large to keep around if they're not being used.

One thing not considered here is composite indexes, which could make sense because we only have a couple distinct queries to worry about. For example, a `[stop_id, vehicle_id]` index might be better than independent `stop_id` and `vehicle_id` indexes, if neither of those are ever queried alone. I think I would like to see how this performs first, though.

losvedir (approved):
👍

PR #323, opened 2020-10-28
chore: extend prune lookback period to keep 7 days of predictions

lemald:
#### Summary of changes
**Asana Ticket:** [📈 capture raw predictions for 1 week](https://app.asana.com/0/584764604969369/1195680449065006/f)

Updated the prune lookback as specified. Per the ticket the pruning should run at least as frequently if not more so, so I doubled it to once every 6 hours instead of once every 12 hours. Looking at recent logs, we haven't come close to hitting our pruning timeout threshold, so I think this should be pretty safe.

#### Reviewer Checklist


digitalcora (approved):
I rate this PR 💯 * 10 * 60 * 60

PR #324, opened 2020-10-28
feat: track and filter by kind of prediction

digitalcora:
**Asana Ticket:** [📈 tracking kind of prediction and being able to filter](https://app.asana.com/0/584764604969369/1187025758962797)

RTR predictions internally have a "kind" which is:

* "at terminal" when the train is waiting to depart from a terminal
* "reverse" when the prediction is on a trip we expect will occur once the train reverses direction at the end of its current trip
* "mid-trip" in all other circumstances

These are not exposed to consumers, but they are mapped to specific values of the `uncertainty` field in stop time updates. Here we reverse that mapping and store/aggregate the "kind" of predictions. This can be null, since the concept is not relevant to Commuter Rail predictions.

#### Reviewer Checklist

### DIFF
--- a/test/prediction_analyzer/prediction_accuracy/prediction_accuracy_test.exs
+++ b/test/prediction_analyzer/prediction_accuracy/prediction_accuracy_test.exs
@@ -17,6 +17,7 @@ defmodule PredictionAnalyzer.PredictionAccuracy.PredictionAccuracyTest do
     route_id: "route1",
     direction_id: 0,
     arrival_departure: "arrival",
+    kind: "mid_trip",
     bin: "0-3 min",
     num_predictions: 10,
     num_accurate_predictions: 5,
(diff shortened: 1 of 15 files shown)


test/prediction_analyzer/prediction_accuracy/prediction_accuracy_test.exs
```diff
@@ -110,6 +111,19 @@ defmodule PredictionAnalyzer.PredictionAccuracy.PredictionAccuracyTest do
...
+      Repo.insert!(%{@prediction_accuracy | kind: nil})
+      Repo.insert!(%{@prediction_accuracy | kind: "at_terminal"})
+      Repo.insert!(%{@prediction_accuracy | kind: "mid_trip"})
+      Repo.insert!(%{@prediction_accuracy | kind: "reverse"})
+
+      {accs, nil} =
+        PredictionAccuracy.filter(Map.merge(base_params(), %{"kinds" => ~w(mid_trip reverse)}))
+
+      query = from(acc in accs, [])
+      assert [%{kind: "mid_trip"}, %{kind: "reverse"}] = execute_query(query)
```
lemald, line 124: Doesn't this rely on the database returning the values in the order they were inserted? Might be better to check that the query returns exactly two values, and one of them has kind `mid_trip` and the other has kind `reverse`, regardless of the order.

lemald:
Since the migrations run asynchronously from the app itself, should we break the migration out into a separate PR?

test/prediction_analyzer/prediction_accuracy/prediction_accuracy_test.exs
digitalcora: It does, but this is in alignment with how most other tests for this module are written, and I didn't think it would cause too much pain — the order should be stable for a given state of the database, and every test starts from a blank slate. If something did cause the order to change, it would be easy to change the tests to match it.

I think it wouldn't complicate things too much to tack on an `order` to the test query, if you think that's worthwhile.

digitalcora:
> Since the migrations run asynchronously from the app itself, should we break the migration out into a separate PR?

Good catch, I will separate them.

test/prediction_analyzer/prediction_accuracy/prediction_accuracy_test.exs
lemald: Ah, yeah, I guess if that's what all the other tests are doing without incident then it's probably fine.

lemald (approved):
👍

digitalcora:
I'll wait on merging this until the RTR change has been deployed to prod, that way we won't have a small span of inconsistent `kind` data where RTR wasn't sending the new value yet.

PR #337, opened 2020-11-04
feat: calculate nth_at_stop when downloading subway predictions

lemald:
#### Summary of changes
**Asana Ticket:** [Collect nth_at_stop data in predictions](https://app.asana.com/0/0/1196327338606456/f)

Followed the choices laid out in the ticket to primarily sort on `arrival_time || departure_time`, additionally sorting predictions with `stops_away` of 0 to the front like realtime signs does. I didn't attempt to replicate any other complicated sign logic.

#### Reviewer Checklist

### DIFF
--- a/lib/fake_httpoison.ex
+++ b/lib/fake_httpoison.ex
@@ -334,7 +334,7 @@ defmodule FakeHTTPoison do
               "id" => "vehicle_id"
             },
             "trip" => %{
-              "trip_id" => "TEST_TRIP",
+              "trip_id" => "TEST_TRIP_1",
               "route_id" => "Red",
               "direction_id" => 1
             },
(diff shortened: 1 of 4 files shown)


lib/prediction_analyzer/predictions/download.ex
```diff
@@ -154,6 +155,17 @@ defmodule PredictionAnalyzer.Predictions.Download do
...
+      |> Enum.map(fn {stop_id, predictions} ->
+        {stop_id,
+         predictions
+         |> Enum.sort_by(
+           &{if(&1.stops_away == 0, do: 0, else: 1), &1.arrival_time || &1.departure_time}
+         )
+         |> Enum.with_index(1)
+         |> Enum.map(fn {prediction, index} -> %{prediction | nth_at_stop: index} end)}
+      end)
+      |> Enum.flat_map(fn {_stop_id, predictions} -> predictions end)
```
digitalcora, line 168: Could we extract this group/map/ungroup operation into a descriptively-named private function like `populate_nth_at_stop`? Maybe then the "main" `flat_map` operation could be moved into a function as well, to keep things at the same level of abstraction.

lib/prediction_analyzer/predictions/download.ex
```diff
@@ -154,6 +155,17 @@ defmodule PredictionAnalyzer.Predictions.Download do
           end)
         end
       end)
+      |> Enum.group_by(& &1.stop_id)
```
digitalcora, line 158: It looks like the `flat_map` that feeds into this can produce nil items in the list (there is no `else` for `if prediction["trip_update"]["stop_time_update"] != nil`). Does this work in that case? Also, does it behave correctly when `stop_id` is nil?

lemald: Huh, good catch. I actually think regardless of my changes the `flat_map` will throw an error if it hits the `nil` case (try `Enum.flat_map([1, 2, 3], fn _x -> nil end)` for instance). I should probably make the function inside the `flat_map` explicitly return an empty list rather than `nil` in that case and test it.

Regarding `nil` stop IDs, I think it would at least not blow up (`Enum.group_by` would work fine), but then you'd have a weird situation where everything with a `nil` stop ID gets ordered together. I suppose the correct behavior there would be to simply not compute `nth_at_stop` for those predictions?

lib/prediction_analyzer/predictions/download.ex
lemald: Yes, sounds like a good idea. I'll do that.

lib/prediction_analyzer/predictions/download.ex
```diff
@@ -162,6 +128,66 @@ defmodule PredictionAnalyzer.Predictions.Download do
...
+        []
+      end
+    end)
+  end
+
+  @spec calculate_nth_at_stop([map()]) :: [map()]
+  defp calculate_nth_at_stop(predictions) do
+    predictions
+    |> Enum.group_by(& &1.stop_id)
+    |> Enum.map(fn {stop_id, predictions} ->
```
digitalcora, line 179: `stop_id` is not used from here on, so instead of

```elixir
|> Enum.map(fn {stop_id, predictions} -> {stop_id, ...transform predictions...})
|> Enum.flat_map(fn {_stop_id, predictions} -> predictions end)
```

could we simplify to:

```elixir
|> Enum.flat_map(fn {_stop_id, predictions} -> ...transform predictions...)
```

lib/prediction_analyzer/predictions/download.ex
digitalcora: I'd agree that we should not populate an `nth_at_stop` for predictions that have no associated stop, since the value will be meaningless.

lib/prediction_analyzer/predictions/download.ex
lemald: Ah, yup, good catch!

lib/prediction_analyzer/predictions/download.ex
lemald: Okay, updated.

lib/prediction_analyzer/predictions/download.ex
```diff
@@ -162,6 +128,68 @@ defmodule PredictionAnalyzer.Predictions.Download do
...
+          end,
+        trip_id: prediction["trip_update"]["trip"]["trip_id"],
+        vehicle_id: prediction["trip_update"]["vehicle"]["id"],
+        route_id: prediction["trip_update"]["trip"]["route_id"],
+        direction_id: prediction["trip_update"]["trip"]["direction_id"],
+        is_deleted: prediction["is_deleted"],
+        nth_at_stop: nil
+      }
+
+      if prediction["trip_update"]["stop_time_update"] != nil do
```
digitalcora, line 149: Come to think of it... in reality, if there are no stop time updates on a prediction, would this just be an empty list, instead of `nil`? Is this conditional always true?

lemald: Yeah I think in practice that's what would happen. On the other hand I'm never against adding handling to make sure a weirdly-formatted input doesn't crash the application, and I've already done so, so I don't see any harm in keeping it in there if you don't.

digitalcora: > Yeah I think in practice that's what would happen.

If we're positive about this, it does seem an oddly specific thing to check for. For example, we don't check for nils before doing `prediction["trip_update"]["vehicle"]["id"]`, which would crash if there wasn't a `vehicle`.

If we're _not_ positive about this, I think we could actually remove the explicit conditional and have the line below be:

```elixir
Enum.reduce(prediction["trip_update"]["stop_time_update"] || [], [], fn update, acc ->
```

lemald: I'm not sure I feel confident enough to remove the handling entirely, but I like your suggestion of rewriting it without the explicit conditional. I'll go ahead and do that.

lib/prediction_analyzer/predictions/download.ex
```diff
@@ -162,6 +128,64 @@ defmodule PredictionAnalyzer.Predictions.Download do
...
+
+  @spec calculate_nth_at_stop([map()]) :: [map()]
+  defp calculate_nth_at_stop(predictions) do
+    predictions
+    |> Enum.group_by(& &1.stop_id)
+    |> Enum.flat_map(fn {stop_id, predictions} ->
+      if !is_nil(stop_id) do
+        predictions
+        |> Enum.sort_by(
+          &{if(&1.stops_away == 0, do: 0, else: 1), &1.arrival_time || &1.departure_time}
```
digitalcora, line 179: Do we need to specially handle predictions that may have neither an arrival nor departure time? For example, predictions involving skipped stops or canceled trips. Since I don't think such a prediction would ever appear on a subway sign, should we give it a nil `nth_at_stop`?

lemald: Ah, yet another good catch. If anything we aren't going to be analyzing those predictions at all, so I think it might be best to simply exclude them from processing further back in `predictions_from_entities/3`.

digitalcora: Works for me, if we are indeed okay with that, though it seems like there might be something there to analyze the accuracy of (e.g. we predicted a vehicle would skip a stop, but it stopped there anyway). I feel like the rest of the pipeline currently isn't set up to do this correctly, though.

lemald: > I feel like the rest of the pipeline currently isn't set up to do this correctly, though.
Yeah, agreed that it would be nice to be able to analyze that sort of thing, but you are correctly that the rest of the app isn't really set up for it right now, so it would be a separate task.

lib/fake_httpoison.ex
```diff
@@ -484,6 +484,34 @@ defmodule FakeHTTPoison do
...
+              "route_id" => "Red",
+              "direction_id" => 1
+            },
+            "stop_time_update" => [
+              %{
+                "arrival" => nil,
+                "boarding_status" => nil,
+                "departure" => nil,
+                "passthrough_time" => 1_540_921_705,
+                "schedule_relationship" => "SCHEDULED",
```
digitalcora, line 508: To be more realistic we could make this `SKIPPED` (I don't think `SCHEDULED`/`nil`/`nil` would ever occur in reality)

digitalcora (approved):
Had one small final (final, final, final) comment. Looks good! 💯

lib/fake_httpoison.ex
lemald: Ah, yup, you're right.

PR #341, opened 2020-11-06
feat: Aggregate nth_at_stop predictions as in_next_two in accuracy

losvedir:
#### Summary of changes
**Asana Ticket:** [Collect in_next_two values in prediction_accuracy](https://app.asana.com/0/584764604969369/1196327338606457)

#### Reviewer Checklist

### DIFF
--- a/lib/prediction_analyzer/prediction_accuracy/query.ex
+++ b/lib/prediction_analyzer/prediction_accuracy/query.ex
@@ -9,6 +9,7 @@ defmodule PredictionAnalyzer.PredictionAccuracy.Query do
           DateTime.t(),
           String.t(),
           String.t(),
+          boolean(),
           String.t(),
           integer(),
           integer(),
@@ -21,6 +22,7 @@ defmodule PredictionAnalyzer.PredictionAccuracy.Query do
         current_time,
         arrival_departure,
         kind,
+        in_next_two?,
         bin_name,
         bin_min,
         bin_max,
@@ -47,7 +49,8 @@ defmodule PredictionAnalyzer.PredictionAccuracy.Query do
       min_unix,
       max_unix,
       environment,
-      kind
+      kind,
+      in_next_two?
     ])
   end
 
@@ -70,6 +73,7 @@ defmodule PredictionAnalyzer.PredictionAccuracy.Query do
         arrival_departure,
         bin,
         kind,
+        in_next_two,
         num_predictions,
         num_accurate_predictions,
         mean_error,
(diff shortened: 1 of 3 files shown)


lib/prediction_analyzer/prediction_accuracy/query.ex
```diff
@@ -110,6 +115,11 @@ defmodule PredictionAnalyzer.PredictionAccuracy.Query do
         AND p.#{arrival_or_departure_time_column} > p.file_timestamp
         AND p.#{arrival_or_departure_time_column} - p.file_timestamp >= $5
         AND p.#{arrival_or_departure_time_column} - p.file_timestamp < $6
+        AND (
```
losvedir, line 118: I ended up going with this approach, since some predictions (i.e. CR) have `nth_at_stop` of `NULL`, I believe. And so I opted to include those values in `in_next_two = FALSE` (so as to not be excluded from accuracy altogether).

That said, it made the query condition for `in_next_two` somewhat complicated. On the one hand, we want `IN (1, 2)` and on the other we want `NOT IN (1, 2) OR IS NULL`. The approach here is the best way I could think of to achieve that relying purely on query parameters, rather than SQL string manipulation (which is, of course, awful).

Once we upgrade to Ecto 3 and remove `arrival` vs `departure`, I wonder if we can actually bring this query back into the realm of Elixir/Ecto. 🤔

lib/prediction_analyzer/prediction_accuracy/query.ex
```diff
@@ -14,13 +14,15 @@ defmodule PredictionAnalyzer.PredictionAccuracy.Query do
           integer(),
           integer(),
           integer(),
-          String.t()
+          String.t(),
+          boolean()
```
lemald, line 18: Looks like the added `boolean()` argument is at the end in the typespec, but the `in_next_two?` argument actually got added more near the middle?

lib/prediction_analyzer/prediction_accuracy/query.ex
lemald: Yeah, makes sense (and yes, `nth_at_stop` will be `NULL` for CR predictions, so I agree with your approach of putting those into `in_next_two = false`).

Redoing this query in Elixir/Ecto would be great! I think this might be the longest string-y SQL query I've seen in a CTD app by a large margin (granted, I haven't looked at any of the fares / AFC 1 stuff...).

lemald (approved):
👍

lemald:
Approved, but see my one minor note about some weirdness with the typespec.

PR #357, opened 2020-12-01
feat: coalesce arrival and departure accuracies

digitalcora:
**Asana Ticket:** [📈 change how we calculate aggregate accuracy](https://app.asana.com/0/584764604969369/1198881667597053)

⬆️ _See ticket comment for comparison of accuracy numbers before and after this change_

Previously, because we separately graded the accuracy of predicted arrival times and departure times, we were "counting" the accuracy of departure times even for predictions that also had arrival times. This didn't align with the reality that if a prediction has an arrival time, *only* that time would be shown to riders on e.g. countdown clocks, and so would be the only time that "mattered". It's only important for a prediction's departure time to be accurate if it doesn't have an arrival time (e.g. terminal departures).

To take this into account in the accuracy calculations, we remove the distinction between `arrival` and `departure` accuracies, and grade a prediction during aggregation as follows:

* if it has an arrival time, compare that to the actual arrival time
* else, if it has a departure time, compare that to the actual departure
* else, don't include in calculations

For now, we're leaving the old accuracies as-is, but the app is no longer aware of the `arrival_departure` field and will always behave as though "All" was selected on the old arrival/departure selector when querying data from before this change.

#### Reviewer Checklist


losvedir (approved):
👍

PR #596, opened 2022-05-04
Deploy to ECS

losvedir:
#### Summary of changes
**Asana Ticket:** Part of [💳 Migrate prediction-analyzer-dev to ECS](https://app.asana.com/0/584764604969369/1199929177825004)

In tandem with switching prediction analyzer to ECS, I had to update the deploy workflow. The most recent GitHub Action "Deploy to ECS" was from this branch, other than the final commit which changed the referenced workflow to `master` in preparation for merge.

One thing I wanted to try out here was using the new [Reusable Workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows). Right now, the `deploy-ecs.yml` workflow calls out to the `ctd-deploy-ecs.yml` workflow in the same directory, but I think it would be great to move `ctd-deploy-ecs.yml` to the mbta/actions (or mbta/workflows?). We could do the same for CI, since almost all apps do the same ASDF / test and lint elixir / optionally test and lint TS thing. And different teams explore the state of the art in terms of concurrency control and structuring those workflows. It would be great if we landed on one set of tooling that all teams could use and then just call out to the shared workflow.

In preparation for using a "reusable workflow" I wanted prediction analyzer to not use distillery, since new CTD apps don't use that anymore. It turns out in #489, we already switched away from distillery, so I added a commit to remove some of its accoutrement. 

#### Reviewer Checklist


losvedir:
New plan per offline discussion: I'll merge this now and we can work on the org-wide reusable workflow after. When it's complete we can open a new PR here to update to use it.

PR #610, opened 2022-05-25
Add matching for Union Square

bklebe:
#### Summary of changes
**Asana Ticket:** [Update RTR to handle Union Sq trains](https://app.asana.com/0/584764604969369/1201366718768696/f)

Added generic_stop_id matching for Union Square, updated Elixir and Erlang versions to latest patch.

#### Reviewer Checklist

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
-elixir 1.10.3-otp-22
-erlang 22.3.4
+elixir 1.10.4-otp-22
+erlang 22.3.4.26
 nodejs 14.17.1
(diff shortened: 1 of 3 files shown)


config/config.exs
```diff
@@ -3,7 +3,7 @@
 #
 # This configuration file is loaded before any dependency and
 # is restricted to this project.
-use Mix.Config
```
losvedir, line 6: I think I remember that if you make this change, you also need to update `Mix.env()` in this file to something else.

bklebe: Good call.

bklebe: I just reverted that because the newer version broke a test I didn't want to fix. Bumped the patch instead

losvedir (approved):

PR #688, opened 2023-01-04
🐛📈 Fix Commuter Rail part of prediction analyzer 

bfauble:
#### Summary of changes
**Asana Ticket:** [🐛📈 Fix Commuter Rail part of prediction analyzer ](https://app.asana.com/0/584764604969369/1203032297707759/f)

kind is nil for commuter rail predictions so they were being ignored for all analyzer data


bklebe (commented):
Do you know whether commuter rail predictions have always had `kind == null` in the database? This seems like it could be a bigger issue.

bklebe (approved):

PR #702, opened 2023-01-30
fix: added missing system packages

nlwstein:
This PR adds a few native libraries missing from the base image that the VM produced errors about missing after upgrading.


paulswartz (approved):
:cake:

*suggestion*: You might also consider adding a line like [this one](https://github.com/mbta/tablespoon/blob/main/Dockerfile#L43-L44) from Tablespoon, which ensures that the release can be started.

PR #766, opened 2023-10-05
wip: reduce file_timestamp buffer for assigning vehicle_ids on glides…

nlwstein:
… events


nlwstein:
@bfauble Deploying this to dev - I can't imagine it will be *worse* than what we are already seeing: https://github.com/mbta/prediction_analyzer/actions/runs/6420767432

PR #774, opened 2023-10-24
feat: 📈  Implement calendar picker

bfauble:
#### Summary of changes
**Asana Ticket:** [📈  Implement calendar picker in Prediction Analyzer](https://app.asana.com/0/584764604969369/1205003249694117/f)

image

* add datepicker to service date, start date, end date fields

### DIFF
--- a/assets/src/chart.ts
+++ b/assets/src/chart.ts
@@ -260,7 +260,7 @@ const setupDatePickers = () => {
       dateSelected: new Date(serviceDateInput.value),
       formatter: dateFormatter,
       showAllDates: true,
-    }).calendarContainer.style.setProperty("font-size", "1.5rem")
+    })
   }
 
   const dateStartInput = document.querySelector(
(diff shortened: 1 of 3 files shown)


assets/src/chart.ts
```diff
@@ -245,7 +246,46 @@ const renderDashboard = () => {
...
+  ) as HTMLInputElement
+  const dateEndInput = document.querySelector(
+    "#filters_date_end"
+  ) as HTMLInputElement
+  if (dateStartInput && dateEndInput) {
+    datepicker(dateStartInput, {
+      dateSelected: new Date(dateStartInput.value),
+      formatter: dateFormatter,
+      showAllDates: true,
+    }).calendarContainer.style.setProperty("font-size", "1.5rem")
```
nlwstein, line 277: Would it be worth doing this through CSS vs. dynamically in JS?

assets/package.json
```diff
@@ -15,6 +15,7 @@
     "@types/jquery": "^3.5.5",
     "c3": "^0.7.20",
     "jquery": "^3.6.0",
+    "js-datepicker": "^5.18.2",
```
nlwstein, line 18: There's a native date picker that you can use as well - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

nlwstein (changes requested):
Just a couple of ideas you may have already considered - otherwise the code looks good to me and I'll approve 🙂

nlwstein (approved):
🚀

PR #775, opened 2023-10-31
[extra] 🐛📈 Prediction Analyzer data selection goes back one day every time the "Filter" button is clicked

bfauble:
#### Summary of changes
**Asana Ticket:** [[extra] 🐛📈 Prediction Analyzer data selection goes back one day every time the "Filter" button is clicked](TICKET_LINK)

* adjust initial date selected based on browser's current offset so that date is not adjusted out of 'correct' date based on timezone.


nlwstein (approved):
🚀

PR #778, opened 2023-11-14
ci: fix asana workflow for dependabot

bklebe:
#### Summary of changes
**Asana Ticket:** [[extra] fix Asana workflow in Prediction Analyzer](https://app.asana.com/0/584764604969369/1205953777846448/f)

Switches to the shared version of the Asana workflow. This should fix the problems with Dependabot and Asana.

#### Reviewer Checklist

### DIFF
--- a/.github/workflows/asana.yml
+++ b/.github/workflows/asana.yml
@@ -2,12 +2,13 @@ name: Asana integration for GitHub PRs
 on: 
   workflow_dispatch:
   pull_request:
-    types: [review_requested, closed]
+    types: [review_requested, closed, opened, reopened]
 
 jobs:
   call-workflow:
     uses: mbta/workflows/.github/workflows/asana.yml@main
     with:
+      development-section: "In Development"
       review-section: "Pending Review"
       merged-section: "Merged / Not Deployed"
       attach-pr: true
--- a/.github/workflows/asana.yml
+++ b/.github/workflows/asana.yml
@@ -1,38 +1,16 @@
-name: Asana integration for Github PRs
+name: Asana integration for GitHub PRs
 on: 
   workflow_dispatch:
   pull_request:
-    types: [closed, edited, opened, reopened]
+    types: [review_requested, closed]
 
 jobs:
-  move-to-merged-asana-ticket-job:
-    runs-on: ubuntu-latest
-    if: github.event.pull_request.merged == true
-    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/asana.yml
```diff
@@ -1,38 +1,16 @@
...
-    steps:
-      - name: Create pull request attachments
-        uses: Asana/create-app-attachment-github-action@latest
-        id: postAttachment
-        with:
-          asana-secret: ${{ secrets.ASANA_GITHUB_INTEGRATION_SECRET }}
-      - name: Log output status
-        run: echo "Status is ${{ steps.postAttachment.outputs.status }}"
+  call-workflow:
+    uses: mbta/workflows/.github/workflows/asana.yml@main
```
meagharty, line 9: It looks like there is also an asana workflow definition in this repo too, can we delete that file?

https://github.com/mbta/prediction_analyzer/blob/master/.github/workflows/asana.yml

.github/workflows/asana.yml
```diff
@@ -1,38 +1,16 @@
...
+    with:
+      review-section: "Pending Review"
+      merged-section: "Merged / Not Deployed"
+      attach-pr: true
```
meagharty, line 13: More of a deploy related question for this project: could we use `merged-section: "Done"` and `complete-on-merge: true`?

meagharty (approved):
LGTM! 

I added a deploy/process related question and a cleanup suggestion.

.github/workflows/asana.yml
bklebe: >It looks like there is also an asana workflow definition in this repo too, can we delete that file?
>
>https://github.com/mbta/prediction_analyzer/blob/master/

If you mean deleting `.github/workflows/asana.yml`, I believe this is the file I edited? Sorry, I'm a bit confused here.

.github/workflows/asana.yml
bklebe: I don't believe we have this written down anywhere (though we should) but we typically don't consider things done until they are deployed to prod.

.github/workflows/asana.yml
meagharty: Woops, ignore this. I was thinking there were two files in this repo, one for a re-usable workflow that wasn't being used anymore and one for this repo's workflow. There's only one for this repo's workflow.

PR #785, opened 2025-01-10
chore: Use TLS-encrypted DB connection

jzimbel-mbta:
#### Summary of changes
**Asana Ticket:** [[extra] 📈 Force TLS for Prediction Analyzer on DB connect](https://app.asana.com/0/584764604969369/1206249244818775/f)

I followed the precedent of other TID projects for this change.

One thing that confused me, though, was the SHA checksum that some of our projects use.

[Skate](https://github.com/mbta/skate/blob/main/Dockerfile#L40-L43) and [Arrow](https://github.com/mbta/arrow/blob/master/Dockerfile#L16-L19) both verify the checksum, but they expect different hashes! I'd expect at least one of these two Docker builds to fail as a result, but they both work. 🤔

Because of that, I left out the checksum verification in this PR. Which is done in at least one other project, [Glides](https://github.com/mbta/glides/blob/main/Dockerfile#L81), so I think it's ok.
Nevertheless I'd be happy to add it if I can get some clarity on what hash I should be verifying against.

#### Reviewer Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,8 @@ WORKDIR /root
 RUN mix local.hex --force && \
   mix local.rebar --force
 
+ADD https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem aws-cert-bundle.pem
+
 ADD mix.lock mix.lock
 ADD mix.exs mix.exs
 ADD config config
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -8,6 +8,8 @@ WORKDIR /root
 RUN mix local.hex --force && \
   mix local.rebar --force
 
+ADD https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem aws-cert-bundle.pem
```
jzimbel-mbta, line 11: Heads up, I left out the checksum verification that some other TID projects do in this step because I was confused about exactly what hash we should be verifying against.

More details in the [PR description](https://github.com/mbta/prediction_analyzer/pull/785#issue-2780344997).

Whoops (approved):

PR #788, opened 2025-03-18
feat(env): add dev-blue

rudiejd:
#### Summary of changes
**Asana Ticket:** [[extra] 📈 Add RTR dev-blue to Prediction Analyzer + minor UI changes](https://app.asana.com/0/584764604969369/1209743714456990/f)
Problem:
We only include data for `rtr-dev-green`, but we'd like to analyze other environments as well

Solution:
Add `rtr-dev-blue` to prediction analyzer

#### Reviewer Checklist

### DIFF
--- a/.envrc.template
+++ b/.envrc.template
@@ -1,5 +1,7 @@
 export AWS_PREDICTIONS_URL=https://s3.amazonaws.com/mbta-gtfs-s3/rtr/TripUpdates_enhanced.json
 export AWS_VEHICLE_POSITIONS_URL=https://s3.amazonaws.com/mbta-gtfs-s3/rtr/VehiclePositions_enhanced.json
-export DATABASE_ROOT_URL=postgres://postgres@localhost:5432
-export DEV_GREEN_AWS_PREDICTIONS_URL=https://s3.amazonaws.com/mbta-gtfs-s3-dev-green/rtr/TripUpdates_enhanced.json
+export DATABASE_ROOT_URL=postgres://postgres@localhost:8083
+export DEV_GREEN_AWS_PREDICTIONS_URL=https://s3.amazonaws.com/mbta-gtfs-s3-dev-gtDen/rtr/TripUpdates_enhanced.json
 export DEV_GREEN_AWS_VEHICLE_POSITIONS_URL=https://s3.amazonaws.com/mbta-gtfs-s3-dev-green/rtr/VehiclePositions_enhanced.json
+export DEV_BLUE_AWS_PREDICTIONS_URL=https://s3.amazonaws.com/mbta-gtfs-s3-dev-blue/rtr/TripUpdates_enhanced.json
+export DEV_BLUE_AWS_VEHICLE_POSITIONS_URL=https://s3.amazonaws.com/mbta-gtfs-s3-dev-blue/rtr/VehiclePositions_enhanced.json
--- a/.envrc.template
+++ b/.envrc.template
@@ -1,7 +1,7 @@
 export AWS_PREDICTIONS_URL=https://s3.amazonaws.com/mbta-gtfs-s3/rtr/TripUpdates_enhanced.json
 export AWS_VEHICLE_POSITIONS_URL=https://s3.amazonaws.com/mbta-gtfs-s3/rtr/VehiclePositions_enhanced.json
 export DATABASE_ROOT_URL=postgres://postgres@localhost:8083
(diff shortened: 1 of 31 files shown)


assets/src/chart.ts
```diff
@@ -178,25 +194,25 @@ const renderDashboard = () => {
...
   const xData = ["x"].concat(sortedBucketNames)
-  let data
+
+  let columns = [xData, col1];
+
   if (showDevGreen) {
-    data = { x: "x", columns: [xData, col1, col2], type: dataType, xFormat }
-  } else {
-    data = { x: "x", columns: [xData, col1], type: dataType, xFormat }
+    let col2: any = ["Dev Green"].concat(sortedDgAccs)
```
rudiejd, line 202: I had to annotate these as `any` to not make typescript explode

assets/src/chart.ts
```diff
@@ -36,6 +37,20 @@ export default () => {
         jQuery("#dev-green-data-table").hide()
         jQuery("#dev-green-accuracy-total").hide()
       }
+      console.log("hello")
```
cmaddox5, line 40: Can this be removed?

assets/src/chart.ts
```diff
@@ -36,6 +37,20 @@ export default () => {
...
+      setupDashboard()
+    })
+
+    jQuery("#show-dev-blue-check").change((event) => {
```
cmaddox5, line 44: What year is it?!

.envrc.template
```diff
@@ -1,5 +1,7 @@
 export AWS_PREDICTIONS_URL=https://s3.amazonaws.com/mbta-gtfs-s3/rtr/TripUpdates_enhanced.json
 export AWS_VEHICLE_POSITIONS_URL=https://s3.amazonaws.com/mbta-gtfs-s3/rtr/VehiclePositions_enhanced.json
-export DATABASE_ROOT_URL=postgres://postgres@localhost:5432
+export DATABASE_ROOT_URL=postgres://postgres@localhost:8083
```
cmaddox5, line 3: Not a port I've seen used with psql before. Was this change intentional?

priv/repo/structure.sql
```diff
@@ -345,5 +346,31 @@ ALTER TABLE ONLY public.predictions
...
+INSERT INTO public."schema_migrations" (version) VALUES (20201019175956);
+INSERT INTO public."schema_migrations" (version) VALUES (20201022151038);
+INSERT INTO public."schema_migrations" (version) VALUES (20201027201115);
+INSERT INTO public."schema_migrations" (version) VALUES (20201103215547);
+INSERT INTO public."schema_migrations" (version) VALUES (20201112161111);
+INSERT INTO public."schema_migrations" (version) VALUES (20201130194309);
+INSERT INTO public."schema_migrations" (version) VALUES (20230717150451);
+INSERT INTO public."schema_migrations" (version) VALUES (20230914092018);
+INSERT INTO public."schema_migrations" (version) VALUES (20231005191442);
+INSERT INTO public."schema_migrations" (version) VALUES (20250318190252);
```
cmaddox5, line 376: It looks like this migration is the one you removed in [this](https://github.com/mbta/prediction_analyzer/pull/788/commits/215dbf832c7bb00fac2b8ca42684b40d78ef7493) commit. The new ENUM type also made it's way in here. Might need to drop and remigrate to update it.

assets/src/chart.ts
rudiejd: lol, yeah

.envrc.template
rudiejd: yes, I just wanted to be able to run arrow DB and this DB at the same time 😅

priv/repo/structure.sql
rudiejd: good call. The new enum type has been added, but I had to get infra to manually run the command due to Postgres issues. With that in mind, do you think I should just remove all references to that migration from the code?

cmaddox5: Hmm, it would be good to have a migration for local development, but I'm not sure how that would work. Will the app work locally without the new enum value?

rudiejd: the app crashes without the enum value. But I could always remove the migration and rely on people loading in the seed data, which adds `dev-blue` to the enum

rudiejd: Added a note to the readme!

cmaddox5 (approved):

PR #789, opened 2025-03-27
fix(chart): date picker is back

rudiejd:
#### Summary of changes
**Asana Ticket:** [[extra] 🐛📈 Fix Prediction Analyzer calendar picker](https://app.asana.com/0/584764604969369/1209814602688445/f)

Problem:
The calendar date picker is not appearing when you click on the service date anymore since we added `dev-blue` as an RTR environment

Solution:
I borked some of the logic for adding the date pickers - they were getting added twice after some changes I made in service of getting the chart and tables ("dashboard") to actually re-render whenever the "show dev green/blue" checkboxes are clicked. I fixed this logic. 

#### Reviewer Checklist

### DIFF
--- a/assets/src/chart.ts
+++ b/assets/src/chart.ts
@@ -16,7 +16,6 @@ declare global {
     sortOrderLink: HTMLElement
     dataPoints: DataPoint[]
     dataPredictionAccuracyJSON: any
-    datePickerCreated: boolean
     chartType: string
     timeframeResolution: string
   }
--- a/assets/src/chart.ts
+++ b/assets/src/chart.ts
@@ -16,6 +16,7 @@ declare global {
     sortOrderLink: HTMLElement
     dataPoints: DataPoint[]
     dataPredictionAccuracyJSON: any
+    datePickerCreated: boolean
     chartType: string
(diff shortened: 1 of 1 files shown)


assets/src/chart.ts
```diff
@@ -16,6 +16,7 @@ declare global {
     sortOrderLink: HTMLElement
     dataPoints: DataPoint[]
     dataPredictionAccuracyJSON: any
+    datePickerCreated: boolean
```
jzimbel-mbta, line 19: I'm confused why you've added this field to the `Window` interface—your code below sets `datePickerAdded` as an attribute on an element within the document.
Should this line be removed?

jzimbel-mbta (approved):
Looks good, just unsure about the change to the `Window` interface.

assets/src/chart.ts
rudiejd: yep, this can be removed!

rudiejd: yep, removed! I was going to do this as a global variable but I decided it was slightly cleaner (and less pollution of the global scope) to do it with an attribute

PR #798, opened 2025-07-31
Leisha Bhandari South Station Prediction Accuracy tracker

leishab:
#### Summary of changes
**Asana Ticket:** [Same-Day Hourly Drop Alert: Prediction accuracy tracker](https://app.asana.com/1/15492006741476/project/584764604969369/task/1210675243067899?focus=true)

- Creatred an alert system to notify when prediction accuracy drops by more than 25% for the terminal "South Station"
- The alert will compare accuracy between consecutive hourly windows, trigger a Slack message to the #tid-transit-data-alerts-testing channel, and include accuracy values and percentage drop. 
- The system should skip alerts if there is no data for the two hours being compared

#### Reviewer Checklist


lib/prediction_analyzer/prediction_accuracy/accuracy_tracker.ex
```diff
@@ -1,104 +1,62 @@
 defmodule PredictionAnalyzer.PredictionAccuracy.AccuracyTracker do
   use GenServer
-
-  alias PredictionAnalyzer.PredictionAccuracy.PredictionAccuracy
-  alias PredictionAnalyzer.Filters
+  alias PredictionAnalyzer.PredictionAccuracy
   require Logger
 
-  @drop_threshold 0.1
+  @target_stop_id "place-south-station"
```
lemald, line 6: **question:** is this supposed to be South Station for Commuter Rail or for the Red Line? The actual stop ID is `place-sstat`, and depending on whether you're looking for CR or Red Line there may be considerations as to whether the parent stop ID is used or a more specific child stop ID referring to the specific platform.

lib/prediction_analyzer/prediction_accuracy/accuracy_tracker.ex
```diff
@@ -1,104 +1,62 @@
...
     schedule_next_check(self())
-    {:ok, [yesterdays_accuracy: 0, previous_days_accuracy: 0]}
-  end
-
-  def check_accuracy() do
-    ["Red", "Blue", "Orange", "Green-B", "Green-C", "Green-D", "Green-E", "Mattapan"]
-    |> Enum.reduce(%{}, fn route_id, acc ->
-      Map.put(acc, route_id, get_route_accuracy(route_id))
-    end)
-  end
```
lemald, line 24: **question:** clarification question on requirements, is the idea that the existing accuracy check is being removed in place of the new one that you're adding?

test/prediction_analyzer/prediction_accuracy/accuracy_tracker_tests.exs
lemald, line 1: **nitpick:** empty file

lemald:
Okay @leishab I have done a full pass of reviewing and left comments, please let me know if you have any questions about my feedback. A couple of other quick thoughts:

1. It looks like there's a conflict with the master branch that should be resolved.
2. It looks like there are also some CI failures; it's probably worth getting those addressed and then trying to deploy this to dev so that we can see it in action.

lib/prediction_analyzer/prediction_accuracy/accuracy_tracker.ex
```diff
@@ -1,104 +1,62 @@
...
+      IO.puts("""
 
-    accuracy
+      Alert!!! Prediction Accuracy drop!!!
+
+      Time: #{inspect(previous_hour_end)} to #{inspect(current_hour_finish)}
+      Previous Hour: #{prev}
+      Current Hour: #{curr}
+      Drop: #{drop * 100}%
+      """)
```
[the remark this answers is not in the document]
lemald: You can always log things to Splunk but not set up an alert just yet; searching for the output in Splunk can also be a good way of getting a hang of how it works before trying to set up an alert.

.DS_Store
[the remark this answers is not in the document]
leishab: I have been trying to find this file, but have had no luck doing so. I searched in the files + tried cd into it but it doesn't exist. Am I doing something wrong here?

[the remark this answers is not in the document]
leishab: Got it let me try that out

lib/prediction_analyzer/prediction_accuracy/accuracy_tracker.ex
leishab: When I talked to Firas, what I took away was I needed to write the code for this from scratch, and not really build upon what already exists. I might be mistaken, but I don't really see a way I can change this code to do what is mentioned in the ticket. Is there another approach. I will keep pushing what I'm doing currently, but if you think there's a different way I should be thinking about this, please let me know!

PR #800, opened 2025-08-19
feat(vehicle_positions/comparator): bespoke max_dwell_time for mattapan

firestack:
#### Summary of changes
**Asana Ticket:** [🔮📈 Investigate and fix "one departure, multiple events" errors at Mattapan (70276)](https://app.asana.com/1/15492006741476/project/1203014709808707/task/1211176719559267?focus=true)

This shortens the amount of time used to look back for previous arrivals when matching departures.

On Mattapan, a train can do a round trip faster than our current `max_dwell_time` of 30m. This means that previous arrivals without departures are then left "eligible" for matching future departures.

[Splunk Dashboard Showing the change of when this branch was deployed between `2025-08-19T15:09:21.392-04:00` and `2025-08-25T10:12:38.972-04:00`](https://mbta.splunkcloud.com/en-US/app/search/transit_datainvestigation_1dmu_max-dwell-time?form.cfg_dd_span=span%3D1h&form.cfg_timechart_span=span%3D123s&form.event-1dmu=One+departure%2C+multiple+updates&form.global_time.earliest=2025-08-18T04%3A00%3A00.000Z&form.global_time.latest=2025-08-26T04%3A00%3A00.000Z&form.pa-indexes=prediction-analyzer-prod&form.pa-indexes=prediction-analyzer-dev&form.usercfg_dd_compare_environment=prod&form.usercfg_ms_environments=prod&form.usercfg_ms_environments=dev-blue&form.usercfg_stop_id=70276&tab=layout_1i7C3Hcc) 

This doesn't address the underlying issue of multiple arrivals or departures not being caught, but it at least reduces the amount of "one departure multiple updates" by up to 50% in non-LRTP environments.

#### Reviewer Checklist

### DIFF
--- a/lib/prediction_analyzer/pruner.ex
+++ b/lib/prediction_analyzer/pruner.ex
@@ -24,10 +24,9 @@ defmodule PredictionAnalyzer.Pruner do
     Logger.info("Beginning prune of DB")
 
     prune_lookback_sec = Application.get_env(:prediction_analyzer, :prune_lookback_sec)
-    max_dwell_time_sec = Application.get_env(:prediction_analyzer, :max_dwell_time_sec)[:default]
 
     predictions_cutoff = System.system_time(:second) - prune_lookback_sec
-    vehicle_events_cutoff = predictions_cutoff - max_dwell_time_sec
+    vehicle_events_cutoff = predictions_cutoff - max_dwell_time_sec()
 
     {time, _} =
       :timer.tc(fn ->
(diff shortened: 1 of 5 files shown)


firestack:
I think you could argue that this both makes the issue a bit more clear (A: makes duplicate arrivals more visible) but also hides the issue (B: 1DMU errors no longer fire on previous arrivals that never got departures matched)

I think the benefit of A outweighs the downside of B, because you can still find them with splunk queries such as
```
index=prediction-analyzer-prod
    vehicle=*
    stop_id=70276
    environment=*
| transaction
  vehicle stop_id environment
  startswith="Inserted vehicle arrival"
  maxpause=10m
| search NOT ("Added departure to vehicle event" OR "One departure, multiple updates")
| transaction vehicle stop_id maxspan=5m
```

lib/prediction_analyzer/pruner.ex
```diff
@@ -24,7 +24,7 @@ defmodule PredictionAnalyzer.Pruner do
     Logger.info("Beginning prune of DB")
 
     prune_lookback_sec = Application.get_env(:prediction_analyzer, :prune_lookback_sec)
-    max_dwell_time_sec = Application.get_env(:prediction_analyzer, :max_dwell_time_sec)
+    max_dwell_time_sec = Application.get_env(:prediction_analyzer, :max_dwell_time_sec)[:default]
```
jzimbel-mbta, line 27: Could you clarify why the default is always used in this case?

If it's because we expect the default value to always be >= the Mattapan-specific value (or potentially other special cases added in the future), would it make sense to use the maximum of all the different values, to be sure that we don't accidentally delete vehicle events that should be kept around?

Basically, I'm wondering if this should be more like
```ex
max_dwell_time_sec =
  Application.get_env(:prediction_analyzer, :max_dwell_time_sec)
  |> Keyword.values()
  |> Enum.max()
```

firestack: Makes sense! How's this? [feedback:fix:doc: make pruner's max_dwell_time_sec use the largest value](https://github.com/mbta/prediction_analyzer/pull/800/commits/2f9d20c6417d9457478a19c66461fc3963e42000)

